PR #8, opened 2023-06-16
Build the basic structure and implement `artiq_client ls` function

### LINKED ISSUE #3: Build the basic structure and implement `artiq_client ls` function
We want to implement two things:
1. Build the basic structure of `FastAPI` server.
2. Implement a proxy to call the `artiq_client ls` command.

The structure will follow the official guide [FastAPI](https://fastapi.tiangolo.com/ko/tutorial/first-steps/).
And the proxy for `artiq_client` command will be implement similarly to [artiq_client.py](https://github.com/m-labs/artiq/blob/master/artiq/frontend/artiq_client.py).

### Additional context
The method how to access the PC which connects to ARTIQ is written in the Obsidian due to the security.

BECATRUE:
This will close #3.

As I wrote in #3, I implemented two things:
1. Build the basic structure of `FastAPI` server.
2. Implement a proxy to call the `artiq_client ls` command.

First of all, if you are not familiar to `FastAPI`, please refer to [tutorial](https://fastapi.tiangolo.com/tutorial/first-steps/).

In the repositoy, there are three files.
- `main.py`: A unique python file. There is everything related to the server.
- `setup.json`: A set-up file similar to one of `iquip`. There is only one element, `master_path`, yet.
- `server.sh`: It makes us run the server easily.

You can run the server as follows:
```shell
$ ./server.sh
```

Also, you can get the list of elements of the specific path.
1. Confirm the master path of `setup.json`.
2. Enter the below command. (Or, you can just enter the url in your browser!)
```shell
$ curl -X GET 127.0.0.1:8000/ls/?directory="blah/blah/"
$ curl -X GET 127.0.0.1:8000/ls/  # default directory path: ""
```
The example result is as follows.


BECATRUE:
If you don't know about `lifespan`, please refer to [this](https://fastapi.tiangolo.com/advanced/events/).

BECATRUE:
And I don't like using a global variable very much.. But, I decided to follow the tutorial first.

main.py
```diff
@@ -0,0 +1,69 @@
...
+    """Loads set-up information from the setup file.
+
+    The file should have the following JSON structure:
+
+      {
+        "master_path": {master_path}
+      }
+    """
+    global settings
+    with open("setup.json", encoding="utf-8") as setup_file:
```
kangz12345: How about calling it "config.json"? It could be confused with "setup.py". (I also want to change the name in `qiwis` 😂)

main.py
```diff
@@ -0,0 +1,69 @@
+"""Proxy server to communicate a client to ARTIQ."""
+
+from contextlib import asynccontextmanager
+import json
+import os
```
kangz12345: We have written `import` lines first, and then `from ... import` lines.

```suggestion
import json
import os
from contextlib import asynccontextmanager
```

main.py
```diff
@@ -0,0 +1,69 @@
...
+
+from fastapi import FastAPI
+from sipyco.pc_rpc import Client
+
+
```
kangz12345, line 10: I think we agreed to put only a single empty line right after the import lines.

main.py
```diff
@@ -0,0 +1,69 @@
...
+@asynccontextmanager
+async def lifespan(_app: FastAPI):
+    """Lifespan events.
+
+    This function is set as the lifespan of the application.
+
+    The code before 'yield' is executed once before the application starts.
+    The code after 'yield' is executed when the application is shutting down.
```
kangz12345: Even though it was helpful for me to understand the code, but I think we don't have to explain this in the code.
Since we use this explicitly `app = FastAPI(lifespan=lifespan)`, anyone who does not know about the lifespan can read the FastAPI documentation.

When you write comments, you may exclude the general information about the API or library. Assume that the reader knows the library as well as you do. Such information can be given in the PR, and it could help the reviews a lot, as you did 👍

main.py
```diff
@@ -0,0 +1,69 @@
...
+def load_setup_file():
+    """Loads set-up information from the setup file.
+
+    The file should have the following JSON structure:
+
+      {
+        "master_path": {master_path}
+      }
+    """
+    global settings
```
kangz12345: As you said, I don't like to use the `global` keyword either.
One workaround would be to pass it as an argument of the function.
Since dictionary is a mutable container, we can [`update()`](https://docs.python.org/3/library/stdtypes.html#dict.update) the contents in the function.

main.py
```diff
@@ -0,0 +1,69 @@
...
+    The code after 'yield' is executed when the application is shutting down.
+    """
+    load_setup_file()
+    yield
+
+
+app = FastAPI(lifespan=lifespan)
+
+
+@app.get("/ls/")
```
kangz12345, line 40: Which is the preferred style: `/ls` vs. `/ls/`? (just curious)

main.py
```diff
@@ -0,0 +1,69 @@
+"""Proxy server to communicate a client to ARTIQ."""
+
+from contextlib import asynccontextmanager
+import json
+import os
+
+from fastapi import FastAPI
+from sipyco.pc_rpc import Client
```
kangz12345: I think `Client` is too general name.. How about importing the module instead?

For example:
```python
from sipyco import pc_rpc
```
or
```python
from sipyco import pc_rpc as rpc
```
if you don't like the underscore in a module name.

main.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+app = FastAPI(lifespan=lifespan)
+
+
+@app.get("/ls/")
+async def list_directory(directory: str = ""):
+    """Get the list of elements in the given path.
+
+    The 'master_path' is used for the prefix of the path.
```
kangz12345: Not a big deal, but `"master_path"` seems more consistent with our quote convention.

kangz12345: In addition, let's clarify that it is the "master_path" in the configuration file.

main.py
```diff
@@ -0,0 +1,69 @@
...
+    """
+    remote = get_client("master_experiment_db")
+    contents = remote.list_directory(os.path.join(settings["master_path"], directory))
+    return contents
```
kangz12345: Hmm, from this code, the return type is quite invisible.
How do we annotate such information for the FastAPI path operation functions?

main.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+@app.get("/ls/")
+async def list_directory(directory: str = ""):
+    """Get the list of elements in the given path.
+
+    The 'master_path' is used for the prefix of the path.
+    """
+    remote = get_client("master_experiment_db")
+    contents = remote.list_directory(os.path.join(settings["master_path"], directory))
```
kangz12345: Could you test if the `AsyncioClient` works as well?
Since we defined this as a coroutine, we can use `await` here.

Even though the FastAPI will run it asynchronously, I think we can practice to use `asyncio`!

In the [reference](https://fastapi.tiangolo.com/async/#in-a-hurry) you provided:

> Anyway, in any of the cases above, FastAPI will still work asynchronously and be extremely fast.

kangz12345 (changes requested):

main.py
BECATRUE: Oh, I agree with you strongly😮 I will change it.

main.py
BECATRUE: I didn't notice it😅 I will rearrange it.

main.py
BECATRUE: Umm.. I'm also curious about it.

In my previous project, our team used the trailing slash so I just use it in this.
As I searched, there is no preferred style and it is up to the developer's preference.

I found two representative standard:
1. [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986#section-3) - "Uniform Resource Identifier (URI): Generic Syntax"
They doesn't use the trailing slash.
2. [W3C](https://www.w3.org/TR/url/)
They use the trailing slash.

So.. I think it is just our preference and I prefer it since the root url also has it. (http://127.0.0.1:8000/)

main.py
BECATRUE: Oh, it's a smart solution! I changed it using `update()` and it still runs correctly.

main.py
BECATRUE: Good! I will apply the latter one.

main.py
BECATRUE: Referred to [Response Model](https://fastapi.tiangolo.com/ko/tutorial/response-model/), we can restrict the return type!
I will apply it in according to the docs.

main.py
BECATRUE: I didn't analyze for the `AsyncioClient`.. And also `artiq_client.py` doesn't use it.
In `artiq_dashboard.py`, it uses both `Client` and `AsyncioClient` so if we need, we can look at it.

If you think we should change into `asyncio`, I'd like to make it as another issue since this proxy server causes a bottlenect of `iquip` project.

main.py
kangz12345: Wow, what a fancy feature!
However, I think the `response_model` feature is not necessary for our case, since we know that the return type is `List[str]`, and it is not even a _Pydantic model_.
Isn't it enough to use just the type annotation?

main.py
```diff
@@ -1,58 +1,57 @@
...
-async def list_directory(directory: str = ""):
-    """Get the list of elements in the given path.
+@app.get("/ls/", response_model=List[str])
+async def list_directory(directory: str = "") -> List[str]:
+    """Get the list of elements in the given path and returns it.
+
+    The "master_path" in the configuration file is used for the prefix of the path.
 
-    The 'master_path' is used for the prefix of the path.
+    Args:
```
kangz12345, line 46: 👍

kangz12345 (changes requested):

main.py
kangz12345: Oh thank you for the research!

main.py
kangz12345: Oh, okay then. Let's use `Client` for now.

main.py
BECATRUE: That's right! I set up the model for future consistency, but I'll take it out if we don't need it.

kangz12345: As far as I understand, we only need it when we use Pydantic models, and want to return in a different type (e.g., in `dict`).
We know that the `remote.list_directory()` returns a list of strings, so we don't need it here.
If we use a Pydantic model later, we can use it that time.

kangz12345 (approved):
LGTM!

PR #10, opened 2023-06-18
Get arguments from the specific experiment

### LINKED ISSUE #5: Implement the function to get arguments from the specific experiment
We want to implement a proxy to get arguments of the specific experiment file.

`artiq.master.experiments.ExperimentDB.examine()` returns everything of the given experiment file.
Using RPC, the server will call this function.

BECATRUE:
This will close #5.

I implemented a proxy to get arguments from the specific experiment.

You can get arguments as follows.
```shell
$ curl -X GET 127.0.0.1:8000/args/?file=exp.py
```

The example result is as follows.

```json
{
  "PulseSimulation": {
    "name": "Pulse simulation",
    "arginfo": {
      "user": [{"ty": "StringValue", "default": "QuIQCL"}, null, null],
      "time": [{"ty": "NumberValue", "default": 1.0, "unit": "", "scale": 1.0, "step": 0.1, "min": null, "max": null, "ndecimals": 2, "type": "auto"}, null, null],
      "save": [{"ty": "BooleanValue", "default": false}, null, null],
      "color": [{"ty": "EnumerationValue", "default": "r", "choices": ["r", "g", "b"]}, null, null]
    },
    "argument_ui": null,
    "scheduler_defaults": {}
  }
}
```

### DIFF
--- a/main.py
+++ b/main.py
@@ -50,6 +50,11 @@ async def list_directory(directory: str = "") -> List[str]:
     return remote.list_directory(os.path.join(configs["master_path"], directory))
 
 
+@app.get("/args/")
+async def get_arguments(file: str):
+    pass
+
+
 def get_client(target_name: str) -> rpc.Client:
     """Creates a client connecting to ARTIQ and returns it.
 
--- a/main.py
+++ b/main.py
@@ -2,6 +2,7 @@
 
 import json
 import os
+import pydantic
 from contextlib import asynccontextmanager
 from typing import Any, Dict, List
 
@@ -50,8 +51,13 @@ async def list_directory(directory: str = "") -> List[str]:
     return remote.list_directory(os.path.join(configs["master_path"], directory))
 
 
-@app.get("/experiment/info/")
-async def get_experiment_info(file: str) -> Dict[str, Dict[str, Any]]:
+class ExperimentInfo(pydantic.BaseModel):
+    name: str
+    arginfo: Dict[str, Any]
+
+
+@app.get("/experiment/info/", response_model=Dict[str, ExperimentInfo])
+async def get_experiment_info(file: str) -> Any:
     """Get information of the given experiment file and returns it.
     
     Args:
@@ -59,9 +65,12 @@ async def get_experiment_info(file: str) -> Dict[str, Dict[str, Any]]:
 
     Returns:
         A dictionary containing only one element of which key is the class name.
-        The value is a dictionary with four keys (show only two important things):
-          name: The experiment name.
+        The value is a dictionary with two keys:
+          name: The experiment name which is set as the docstring in the experiment file.
           arginfo: The dictionary containing arguments of the experiment.
+            Each key is an argument name and its value contains the argument type,
+            the default value, and the additional information for the argument.
+
     """
     remote = get_client("master_experiment_db")
     return remote.examine(file)
--- a/main.py
+++ b/main.py
@@ -3,7 +3,7 @@
 import json
 import os
 from contextlib import asynccontextmanager
-from typing import List
+from typing import Any, Dict, List
 
 from fastapi import FastAPI
 from sipyco import pc_rpc as rpc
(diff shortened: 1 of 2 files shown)


BECATRUE:
The example experiment file `pulse.py` is as follows:
```python

class PulseSimulation(EnvExperiment):
    """Pulse simulation"""

    def build(self):
        self.setattr_device("core")
        self.setattr_device("a")

        self.setattr_argument("user", StringValue(default="QuIQCL"))
        self.setattr_argument("time", NumberValue(default=1))

        self.setattr_argument("save", BooleanValue(default=False))
        self.setattr_argument("color", EnumerationValue(choices=["r", "g", "b"], default="r"))

    @kernel
    def run(self):
        self.a.pulse(100*MHz, self.time*us)

    def analyze(self):
        self.set_dataset("user", self.user, broadcast=True)
```

main.py
```diff
@@ -50,6 +50,23 @@ async def list_directory(directory: str = "") -> List[str]:
     return remote.list_directory(os.path.join(configs["master_path"], directory))
 
 
+@app.get("/args/")
+async def get_arguments(file: str) -> Dict[str, Dict[str, Any]]:
+    """Get arguments of the given experiment file and returns it.
```
kangz12345: The function name and docstring say that this returns the arguments of the given experiment, but it actually returns more than that.
Is the extra information necessary? (For example, experiment class name, argument ui, scheduler defaults.)
Only two (name and arginfo) are described in the docstring.

I think you should choose one of the following:
1. Only return the arginfo.
2. Change the function name to, for example, `get_experiment_info()`.

Moreover, if you want to filter the `argument_ui` and `scheduler_defaults` out, you can use the response model here, I guess.
For example:
```python
import pydantic
class ExperimentInfo(pydantic.BaseModel):
  name: str
  arginfo: Dict[str, Any]

@app.get("/expinfo/", response_model=Dict[str, ExperimentInfo])
async def get_experiment_info(file: str) -> Any:
  remote = get_client(...)
  return remote.examine(file)
```

kangz12345: Moreover, do you know why each argument is given in a list with the trailing `, null, null`?

kangz12345 (changes requested):

main.py
BECATRUE: Oh, I was also thinking about the problem.

I think the experiment name is useful to show it in the builder GUI.
For your information, we can set the name as the first line of the docstring in the experiment class.
Please refer to the above code.

Then, I agree with you to change the function name to `get_experiment_info()` and filter the unnecessary data.
Thanks for suggesting the code😃

BECATRUE: > Moreover, do you know why each argument is given in a list with the trailing `, null, null`?

Honestly, I don't know what they mean.. I tried to find it, but lots of codes are dependent so I didn't😥

kangz12345: Okay then! Please make the docstring clearer (describe the form of the contents in more detail).

BECATRUE: And also, how about changing the routing path to `/experiment/info`?
I want to add the prefix `/experiment/` to all functions about experiments such as `submit`.

kangz12345: That's a great idea!

main.py
```diff
@@ -50,6 +51,31 @@ async def list_directory(directory: str = "") -> List[str]:
...
+
+@app.get("/experiment/info/", response_model=Dict[str, ExperimentInfo])
+async def get_experiment_info(file: str) -> Any:
+    """Get information of the given experiment file and returns it.
+    
+    Args:
+        file: The path of the experiment file.
+
+    Returns:
+        A dictionary containing only one element of which key is the class name.
```
kangz12345: How about "the experiment class name" instead of just "the class name"?

main.py
```diff
@@ -50,6 +51,31 @@ async def list_directory(directory: str = "") -> List[str]:
...
+        The value is a dictionary with two keys:
+          name: The experiment name which is set as the docstring in the experiment file.
+          arginfo: The dictionary containing arguments of the experiment.
+            Each key is an argument name and its value contains the argument type,
+            the default value, and the additional information for the argument.
```
kangz12345: Thank you for the detailed explanation! However, I think it should be in the `ExperimentInfo` class docstring, not here.

kangz12345 (changes requested):
Could you check if it returns the information without `argument_ui` and `scheduler_defaults`?

BECATRUE:
> Could you check if it returns the information without `argument_ui` and `scheduler_defaults`?

Yes! As we intended, it returns only two elements.

main.py
BECATRUE: Oh, it looks better! I will move it.

BECATRUE:
I added the pylintrc file to pass a warning in terms of `pydantic`.
For details, see this [issue](https://github.com/pydantic/pydantic/issues/1961).

kangz12345 (approved):
LGTM!

PR #11, opened 2023-06-18
Submit the specific experiment

### LINKED ISSUE #6: Implement `artiq_client submit` function
What is the feature you want to implement?

ex) Add *B* method to *A* class, Upgrade *C* model

### Additional context
Add any other context or screenshots about the feature here.

BECATRUE:
This will close #6.

I implmented a proxy to submit an experiment.

You can submit it as follows:
```shell
$ http://127.0.0.1:8000/experiment/submit/?file=pulse.py  # no arguments
$ http://127.0.0.1:8000/experiment/submit/?file=pulse.py&args={"time":3,"user":"jaehun"}
```

I thought a lot about how to pass the arguments. The solutions are as follows:
1. Use dynamic parameters
Later, there may be additional parameters such as a due date, a priority.
Thus, it makes us difficult to distinguish the experiment argumetns and these other parameters. 
2. Use a JSONifiable string (_Selected_)
Pass the arguments as a JSONifiable string such as `'{"key1": "value1", "key2": "value2"}'`.
I think it is the most convenient since we can encode and decode with just one function i.e. `json.dumps()` and `json.loads()`.

And I didn't check if a decode exception occurs because it can be handled internally.
If so, the server returns a string "Internal Server Error".

Also regardless of whether the submission is successful or not, the return is always the rid number.
The success or failure can be checked only in the `artiq_master` terminal.

### DIFF
--- a/main.py
+++ b/main.py
@@ -2,6 +2,7 @@
 
 import json
 import os
+import logging
 from contextlib import asynccontextmanager
 from typing import Any, Dict, List
 
@@ -80,9 +81,18 @@ async def get_experiment_info(file: str) -> Any:
     remote = get_client("master_experiment_db")
     return remote.examine(file)
 
+
 @app.get("/experiment/submit/")
-async def submit_experiment(file: str) -> Any:
-    pass
+async def submit_experiment(file: str, args: str = "{}") -> int:
+    """Submit the given experiment file."""
+    expid = {
+        "log_level": logging.WARNING,
+        "class_name": None,
+        "arguments": json.loads(args),
+        "file": file
+    }
+    remote = get_client("master_schedule")
+    return remote.submit("main", expid, 0, None, False)
 
 
 def get_client(target_name: str) -> rpc.Client:
--- a/main.py
+++ b/main.py
@@ -1,7 +1,7 @@
 """Proxy server to communicate a client to ARTIQ."""
 
 import json
-import os
+import posixpath
 import logging
 from contextlib import asynccontextmanager
 from typing import Any, Dict, List
@@ -50,7 +50,7 @@ async def list_directory(directory: str = "") -> List[str]:
         directory: The path of the directory to search for.
     """
     remote = get_client("master_experiment_db")
-    return remote.list_directory(os.path.join(configs["master_path"], directory))
+    return remote.list_directory(posixpath.join(configs["master_path"], directory))
 
 
 class ExperimentInfo(pydantic.BaseModel):
@@ -99,7 +99,7 @@ async def submit_experiment(file: str, args: str = "{}") -> int:
         "log_level": logging.WARNING,
         "class_name": None,
         "arguments": json.loads(args),
-        "file": os.path.join(configs["repository_path"], file)
+        "file": posixpath.join(configs["repository_path"], file)
     }
     remote = get_client("master_schedule")
     return remote.submit("main", expid, 0, None, False)
--- a/main.py
+++ b/main.py
@@ -80,6 +80,10 @@ async def get_experiment_info(file: str) -> Any:
     remote = get_client("master_experiment_db")
     return remote.examine(file)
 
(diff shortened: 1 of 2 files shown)


BECATRUE:
For details about submission arguments, please refer to [artiq_client](https://github.com/m-labs/artiq/blob/master/artiq/frontend/artiq_client.py#L135).

main.py
```diff
@@ -81,6 +83,28 @@ async def get_experiment_info(file: str) -> Any:
     return remote.examine(file)
 
 
+@app.get("/experiment/submit/")
+async def submit_experiment(file: str, args: str = "{}") -> int:
+    """Submit the given experiment file.
+    
+    Args:
+        file: The path of the experiment file.
+        args: The arguments to submit which must be JSONifiable to a dictionary.
```
kangz12345: Usually we say "JSON string", while "JSONifiable" refers to the objects which can be converted into a JSON, and JSON is a well-formatted _string_.
Therefore, "... must be a JSON string of a dictionary." would be more natural!

main.py
```diff
@@ -81,6 +83,28 @@ async def get_experiment_info(file: str) -> Any:
     return remote.examine(file)
 
 
+@app.get("/experiment/submit/")
+async def submit_experiment(file: str, args: str = "{}") -> int:
+    """Submit the given experiment file.
```
kangz12345: Submits!

main.py
```diff
@@ -81,6 +83,28 @@ async def get_experiment_info(file: str) -> Any:
...
+        file: The path of the experiment file.
+        args: The arguments to submit which must be JSONifiable to a dictionary.
+          Each key is an argument name and its value is the value of the argument.
+    
+    Returns:
+        The run identifier, an integer which is incremented at each experiment submission.
+    """
+    expid = {
+        "log_level": logging.WARNING,
+        "class_name": None,
```
kangz12345, line 100: Is this only necessary when there are more than one experiment class in a single file?

main.py
```diff
@@ -81,6 +83,28 @@ async def get_experiment_info(file: str) -> Any:
...
+        "arguments": json.loads(args),
```
kangz12345, line 101: I think it is fine for now, but are the scan variables JSON-native objects as well?
Maybe someday we might want to use non-JSON-native objects, just like in `qiwiscall`s.
Nevertheless, I think we don't have to cover it in this PR.

main.py
```diff
@@ -81,6 +83,28 @@ async def get_experiment_info(file: str) -> Any:
...
+        "file": os.path.join(configs["repository_path"], file)
```
kangz12345: Let's use the `posixpath` here as well!

kangz12345 (changes requested):

main.py
BECATRUE: Oh, thanks for letting me know! I thought I was confused.

main.py
BECATRUE: Wow, all the other functions were written in singular form. I will edit these, too.

main.py
BECATRUE: As I understand, it is correct!

In [tools.py](https://github.com/m-labs/artiq/blob/454597915af820c94f8828a9e6cbcba03a03e600/artiq/tools.py#L94), I expect that it is used to find one of several experiments.

main.py
BECATRUE: I also agree with you.
Yet, I don't know the format of scan variables well, so I couldn't cover it in this PR.
Thus, as you said, it is okay to implement later!

main.py
BECATRUE: Good! I will change one of `list_directory()`, too.

main.py
kangz12345: Thank you for the reference!

kangz12345 (approved):
LGTM

PR #12, opened 2023-06-20
Unit tests

### LINKED ISSUE #7: Unit tests for the implemented functions
What is the feature you want to implement?

ex) Add *B* method to *A* class, Upgrade *C* model

### Additional context
Add any other context or screenshots about the feature here.

BECATRUE:
This will close #7.

I implemented all tests for all functions implemented so far.

You can run the test as follows:
```shell
$ python -m coverage run -m unittest test.py
$ python -m coverage report main.py
```

> This is NOT completed, just to be confirmed about #7.
> 
> I implemented the unit test for `list_directory()` similar to `artiq_client ls`.
> 
> First of all, the test for FastAPI is referred to [Testing](https://fastapi.tiangolo.com/tutorial/testing/).
> In the test, I want only the return value and routing queries since the lifespan and `get_client()` are outside of this function.
> 
> So, I need a review of this structure.

### DIFF
--- a/test.py
+++ b/test.py
@@ -21,12 +21,13 @@ class RoutingTest(unittest.TestCase):
     @mock.patch.dict("main.configs", {"master_path": "master_path/"})
     def test_list_directory(self):
         test_list = ["dir1/", "dir2/", "file1.py", "file2.py"]
-        self.mocked_get_client.return_value.list_directory.return_value = test_list
+        mocked_client = self.mocked_get_client.return_value
+        mocked_client.list_directory.return_value = test_list
         with TestClient(main.app) as client:
             for query, path in [("", ""), ("?directory=dir1/", "dir1/")]:
                 response = client.get("/ls/" + query)
                 self.mocked_load_config_file.assert_called()
-                self.mocked_get_client.return_value.list_directory.assert_called_with(
+                mocked_client.list_directory.assert_called_with(
                     "master_path/" + path
                 )
                 self.assertEqual(response.status_code, 200)
--- a/test.py
+++ b/test.py
@@ -7,7 +7,7 @@ from fastapi.testclient import TestClient
 import main
 
 
-class Test(unittest.TestCase):
+class RoutingTest(unittest.TestCase):
     """Unit tests for routing and each operation."""
 
     def setUp(self):
--- a/test.py
+++ b/test.py
@@ -11,9 +11,26 @@ class Test(unittest.TestCase):
     """Unit tests for routing and each operation."""
 
     def setUp(self):
-        patcher = mock.patch("main.get_client")
-        self.mocked_get_client = patcher.start()
-        self.addCleanup(patcher.stop)
+        patcher_load_config_file = mock.patch("main.load_config_file")
+        patcher_get_client = mock.patch("main.get_client")
+        self.mocked_load_config_file = patcher_load_config_file.start()
+        self.mocked_get_client = patcher_get_client.start()
+        self.addCleanup(patcher_load_config_file.stop)
+        self.addCleanup(patcher_get_client.stop)
+
+    @mock.patch.dict("main.configs", {"master_path": "master_path/"})
+    def test_list_directory_return(self):
+        """Tests if the return value is correct."""
+        test_list = ["dir1/", "dir2/", "file1.py", "file2.py"]
+        self.mocked_get_client.return_value.list_directory.return_value = test_list
+        with TestClient(main.app) as client:
+            response = client.get("/ls/")
+        self.mocked_load_config_file.assert_called_once()
+        self.mocked_get_client.return_value.list_directory.assert_called_once_with(
+            "master_path/"
+        )
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.json(), test_list)
 
 
 if __name__ == "__main__":
--- a/test.py
+++ b/test.py
@@ -34,6 +34,26 @@ class RoutingTest(unittest.TestCase):
                 self.assertEqual(response.status_code, 200)
                 self.assertEqual(response.json(), test_list)
 
+    def test_get_experiment_info(self):
+        test_info = {
+            "ExperimentClass": main.ExperimentInfo(
+                name="experiment_name",
+                arginfo={
+                    "arg1": [{"ty": "StringValue", "default": "DEFAULT"}, None, None],
+                    "arg2": [{"ty": "BooleanValue", "default": True}, None, None]
+                }
+            )
+        }
+        mocked_client = self.mocked_get_client.return_value
+        mocked_client.examine.return_value = test_info
+        with TestClient(main.app) as client:
+            self.mocked_load_config_file.assert_called_once()
+            response = client.get("/experiment/info/?file=experiment.py")
+            mocked_client.examine.assert_called_with("experiment.py")
+            self.assertEqual(response.status_code, 200)
+            self.assertEqual(response.json(), test_info)
+
+
 
 if __name__ == "__main__":
     unittest.main()
--- a/test.py
+++ b/test.py
@@ -1,0 +1,15 @@
+"""Unit tests for main module."""
+
+import unittest
+from unittest import mock
+from fastapi.testclient import TestClient
+
+import main
+
+
+class Test(unittest.TestCase):
+    """Unit tests for routing and each operation."""
+
+
+if __name__ == "__main__":
+    unittest.main()
--- a/test.py
+++ b/test.py
@@ -19,18 +19,18 @@ class Test(unittest.TestCase):
         self.addCleanup(patcher_get_client.stop)
 
     @mock.patch.dict("main.configs", {"master_path": "master_path/"})
-    def test_list_directory_return(self):
-        """Tests if list_directory() returns correctly."""
+    def test_list_directory(self):
         test_list = ["dir1/", "dir2/", "file1.py", "file2.py"]
         self.mocked_get_client.return_value.list_directory.return_value = test_list
         with TestClient(main.app) as client:
-            response = client.get("/ls/")
-        self.mocked_load_config_file.assert_called_once()
-        self.mocked_get_client.return_value.list_directory.assert_called_once_with(
-            "master_path/"
-        )
-        self.assertEqual(response.status_code, 200)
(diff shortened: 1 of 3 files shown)


test.py
```diff
@@ -0,0 +1,37 @@
...
+        patcher_get_client = mock.patch("main.get_client")
+        self.mocked_load_config_file = patcher_load_config_file.start()
+        self.mocked_get_client = patcher_get_client.start()
+        self.addCleanup(patcher_load_config_file.stop)
+        self.addCleanup(patcher_get_client.stop)
+
+    @mock.patch.dict("main.configs", {"master_path": "master_path/"})
+    def test_list_directory(self):
+        test_list = ["dir1/", "dir2/", "file1.py", "file2.py"]
+        self.mocked_get_client.return_value.list_directory.return_value = test_list
```
kangz12345: Defining `mocked_client = self.mocked_get_client.return_value` could be helpful.

test.py
```diff
@@ -0,0 +1,37 @@
...
+        with TestClient(main.app) as client:
+            for query, path in [("", ""), ("?directory=dir1/", "dir1/")]:
+                response = client.get("/ls/" + query)
```
kangz12345: Doesn't the f-string look better?
```python
f"/ls/{query}"
```

test.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def setUp(self):
+        patcher_load_config_file = mock.patch("main.load_config_file")
+        patcher_get_client = mock.patch("main.get_client")
+        self.mocked_load_config_file = patcher_load_config_file.start()
+        self.mocked_get_client = patcher_get_client.start()
+        self.addCleanup(patcher_load_config_file.stop)
+        self.addCleanup(patcher_get_client.stop)
+
+    @mock.patch.dict("main.configs", {"master_path": "master_path/"})
```
kangz12345: 👍

test.py
```diff
@@ -0,0 +1,37 @@
...
+    def test_list_directory(self):
+        test_list = ["dir1/", "dir2/", "file1.py", "file2.py"]
+        self.mocked_get_client.return_value.list_directory.return_value = test_list
+        with TestClient(main.app) as client:
+            for query, path in [("", ""), ("?directory=dir1/", "dir1/")]:
+                response = client.get("/ls/" + query)
+                self.mocked_load_config_file.assert_called()
```
kangz12345: Should we check this for each query?

test.py
```diff
@@ -0,0 +1,37 @@
...
+                self.mocked_get_client.return_value.list_directory.assert_called_with(
+                    "master_path/" + path
+                )
```
kangz12345: We can also write:
```python
mocked_client.list_directory.assert_called_with(f"master_path/{path}")
```

test.py
```diff
@@ -0,0 +1,37 @@
+"""Unit tests for main module."""
+
+import unittest
+from unittest import mock
+from fastapi.testclient import TestClient
+
+import main
+
+
```
kangz12345: Just a single empty line 😂

kangz12345 (changes requested):
The structure seems good enough!

test.py
BECATRUE: Oh.. I'm sorry I missed it too much😥
Even, I found that I made the same mistake in [test_explorer.py](https://github.com/snu-quiqcl/iquip/blob/develop/tests/test_explorer.py).

test.py
BECATRUE: Oh, it looks better!

test.py
BECATRUE: You're right! I will move it out of the loop.

test.py
```diff
@@ -0,0 +1,101 @@
+"""Unit tests for main module."""
+
+import logging
+import json
+import posixpath
+import unittest
+from unittest import mock
+from fastapi.testclient import TestClient
```
kangz12345, line 8: Let's put an empty line between the python native modules and the external 3rd party modules!

test.py
```diff
@@ -0,0 +1,101 @@
...
+    def setUp(self):
+        patcher_load_config_file = mock.patch("main.load_config_file")
+        patcher_get_client = mock.patch("main.get_client")
+        self.mocked_load_config_file = patcher_load_config_file.start()
+        self.mocked_get_client = patcher_get_client.start()
+        self.addCleanup(patcher_load_config_file.stop)
+        self.addCleanup(patcher_get_client.stop)
+
+    @mock.patch.dict("main.configs",
+                     {"master_path": "master_path/", "repository_path": "repo_path/"})
```
kangz12345, line 25: Hmm.. multiline decorator looks strange 😂
How about this:
```suggestion
    _mocked_configs = {
        "master_path": "master_path/",
        "repository_path": "repo_path/",
    }
    @mock.patch.dict("main.configs", _mocked_configs)
```

However, either one is okay for me! Just use what you want.

test.py
```diff
@@ -0,0 +1,101 @@
...
+    def test_list_directory(self):
+        test_list = ["dir1/", "dir2/", "file1.py", "file2.py"]
+        mocked_client = self.mocked_get_client.return_value
```
kangz12345: Maybe this one can be defined in the `setUp()`, i.e., `self.mocked_client`?

test.py
```diff
@@ -0,0 +1,101 @@
...
+        mocked_client.list_directory.return_value = test_list
+        with TestClient(main.app) as client:
+            self.mocked_load_config_file.assert_called_once()
+            for params in [{}, {"directory": "dir1/"}]:
+                response = client.get(f"/ls/", params=params)
```
kangz12345: You can remove the `f` since it is now a pure string literal.

test.py
```diff
@@ -0,0 +1,101 @@
...
+                mocked_client.list_directory.assert_called_with(
+                    f"master_path/repo_path/{params.get('directory', '')}")
```
kangz12345: It looks not readable enough. Let's separate it:
```python
directory = params.get("directory", "")
...
f"master_path/repo_path/{directory}"
```

test.py
```diff
@@ -0,0 +1,101 @@
...
+            )
+        }
+        mocked_client = self.mocked_get_client.return_value
+        mocked_client.examine.return_value = test_info
+        with TestClient(main.app) as client:
+            self.mocked_load_config_file.assert_called_once()
+            response = client.get("/experiment/info/", params={'file': 'experiment.py'})
+            mocked_client.examine.assert_called_with("experiment.py")
+            self.assertEqual(response.status_code, 200)
+            self.assertEqual(response.json(), test_info)
```
kangz12345, line 55: How does this work? `test_info` is not a JSON object since it contains a `main.ExperimentInfo` instance.

test.py
```diff
@@ -0,0 +1,101 @@
...
+    @mock.patch.dict("main.configs", {"repository_path": "repo_path/"})
+    def test_submit_experiment(self):
+        test_rid = 0
+        mocked_client = self.mocked_get_client.return_value
+        mocked_client.submit.return_value = test_rid
+        with TestClient(main.app) as client:
+            self.mocked_load_config_file.assert_called_once()
+            for params in [
+                {"file": "experiment1.py"},
+                {"file": "experiment2.py", "args": '{"k": "v"}'}]:
```
kangz12345: Let's define a variable for better readability:
```python
test_params = [
    {"file": ...},
    {...},
]
for params in test_params:
    ...
```

kangz12345: In addition, I prefer to use tuples instead of lists since they are immutable.

test.py
```diff
@@ -0,0 +1,101 @@
...
+                response = client.get("/experiment/submit/", params=params)
+                mocked_client.submit.assert_called_with(
+                    "main", {
+                        "log_level": logging.WARNING,
+                        "class_name": None,
+                        "arguments": json.loads(params.get("args", "{}")),
+                        "file": posixpath.join("repo_path", params["file"])
+                    }, 0, None, False
```
kangz12345: The code layout seems inconsistent.. Let's define the large dict as a variable or break the line after each argument.

test.py
```diff
@@ -0,0 +1,101 @@
...
+                self.assertEqual(response.status_code, 200)
+                self.assertEqual(response.json(), test_rid)
+
+
+class FunctionTest(unittest.TestCase):
+    """Unit tests for other functions."""
+
+    @mock.patch("builtins.open")
+    @mock.patch("json.load",
+                return_value={"master_path": "master_path/", "repository_path": "repo_path/"})
```
kangz12345, line 85: Choose a style based on your preference and follow the same style as above.

kangz12345 (changes requested):

test.py
BECATRUE: I also thought about your suggestion, but I thought it makes us difficult to understand each test function because we cannot see how is mocked at a glance.

So if you are okay, I want to remain it as it is.

test.py
BECATRUE: Yes! I will move it😃

test.py
BECATRUE: Thanks for suggesting an example!

test.py
BECATRUE: Umm... I searched a lot.. but I couldn't know why it works obviously.
In this [community](https://lightrun.com/answers/pydantic-pydantic-improve-basemodel__eq__-performance), someone said the comparision is operated after converting a model to a dict, and I also think so.

But there is no official reference..

kangz12345:
Is this updated? I don't see the difference.

test.py
kangz12345: That's weird.. anyway, it seems that the comparison is implemented so we could use it.

BECATRUE:
> Is this updated? I don't see the difference.

I updated now.. Sorry😅

test.py
kangz12345: I think you missed this!

BECATRUE:
@kangz12345 I tag you directly because I can't request reviews in the reviewer section😅

test.py
kangz12345: @BECATRUE

BECATRUE: Thanks for reminding me! I will check it now.

BECATRUE: I applied it! @kangz12345

kangz12345 (approved):
LGTM

PR #15, opened 2023-06-20
Modify the default path of `list_directory()`

### LINKED ISSUE #14: Modify the default path of `list_directory()`
Now, the default path of `list_directory()` is set as where `artiq_master` is running, not in the `repository_path`.
But, we use this function to find experiment files, so I think it is more natural to set as the `repository_path`.

BECATRUE:
This will close #14.

As I wrote in #14, it may be more convenient to modify the default path to the `repository_path`.

### DIFF
--- a/main.py
+++ b/main.py
@@ -44,13 +44,16 @@ app = FastAPI(lifespan=lifespan)
 async def list_directory(directory: str = "") -> List[str]:
     """Gets the list of elements in the given path and returns it.
 
-    The "master_path" in the configuration file is used for the prefix of the path.
+    The "master_path" and "repository_path" in the configuration file 
+    is used for the prefix of the path.
 
     Args:
         directory: The path of the directory to search for.
     """
     remote = get_client("master_experiment_db")
-    return remote.list_directory(posixpath.join(configs["master_path"], directory))
+    return remote.list_directory(posixpath.join(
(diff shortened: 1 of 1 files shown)


main.py
```diff
@@ -44,13 +44,16 @@ async def lifespan(_app: FastAPI):
...
+    The "master_path" and "repository_path" in the configuration file 
+    is used for the prefix of the path.
 
     Args:
         directory: The path of the directory to search for.
     """
     remote = get_client("master_experiment_db")
-    return remote.list_directory(posixpath.join(configs["master_path"], directory))
+    return remote.list_directory(posixpath.join(
+        configs["master_path"], configs["repository_path"], directory
```
kangz12345, line 55: Is the repository path always in the master path?
If not, I think we don't have to put them together.
What would you prefer: absolute vs. relative path for the repository path?

BECATRUE: I don't know if the repository path must be in the master path, but the path passed as the `repository` argument of `artiq_master` is defined as a relative path to the master path!
I inteded that the server manager set `repository_path` the same as the argument.

Please refer to [artiq_master](https://m-labs.hk/artiq/manual/management_system.html#repository).

kangz12345: Oh I see, then let's stick to the current implementation.

kangz12345 (approved):
LGTM

PR #40, opened 2023-08-02
Support the visualization option in submission

### LINKED ISSUE #36: Support the visualization option in submission
- Support the visualization option.
- If the option is set, then the experiment code will be copied and saved into the result directory.

- In `submit_experiment()`, add a `visualize` option, which modify the experiment code and save it.
- Save the origin experiment code into the visualize path, which is set in `config.json`.

BECATRUE:
This PR will close #36.

I implemented the followings:
- Remove fetching a specific experiment code (we will not use anymore.)
- Add the visualize option. (The default is not to visualize.)
- If the option is `True`, the original code is copied in `${master_path}/{visualize_path}/{rid}/experiment.py`.

### Result
I sent the query below.
```shell
http://127.0.0.1:8000/experiment/submit/?file=simple_simulation.py&visualize=True
```

Its RID was 277 and in my ARTIQ master path, the following structure was created.

<image width="50%" src="https://github.com/snu-quiqcl/artiq-proxy/assets/76851886/e4ce9366-6e51-4596-a54c-361f07f552c4">

### DIFF
--- a/main.py
+++ b/main.py
@@ -114,7 +114,8 @@ async def submit_experiment(
     args: str = "{}",
     pipeline: str = "main",
     priority: int = 0,
-    timed: Optional[str] = None
+    timed: Optional[str] = None,
+    visualize: Optional[bool] = False
 ) -> int:
     """Submits the given experiment file.
     
@@ -126,10 +127,16 @@ async def submit_experiment(
         priority: Higher value means sooner scheduling.
         timed: The due date for the experiment in ISO format.
           None for no due date.
+        visualize: If True, the experiment file is modified for visualization.
+          The original file and vcd file are saved in the visualize path set in config.json.
     
     Returns:
         The run identifier, an integer which is incremented at each experiment submission.
     """
+    if visualize:
+        pass
+    else:
+        pass
     expid = {
         "log_level": logging.WARNING,
         "class_name": None,
--- a/main.py
+++ b/main.py
@@ -95,19 +95,6 @@ async def get_experiment_info(file: str) -> Any:
     return remote.examine(file)
 
 
-@app.get("/experiment/code/")
-async def get_experiment_code(file: str) -> str:
-    """Gets code of the given experiment file and returns it.
-    
-    Args:
-        file: The path of the experiment file.
-    """
-    full_path = posixpath.join(configs["master_path"], configs["repository_path"], file)
-    with open(full_path, encoding="utf-8") as experiment_file:
-        code = experiment_file.read()
-    return code
-
-
 @app.get("/experiment/submit/")
 async def submit_experiment(
     file: str,
--- a/main.py
+++ b/main.py
@@ -121,7 +121,10 @@ async def submit_experiment(
         The run identifier, an integer which is incremented at each experiment submission.
     """
     if visualize:
-        pass
+        full_path = posixpath.join(configs["master_path"], configs["repository_path"], file)
+        with open(full_path, encoding="utf-8") as experiment_file:
+            code = experiment_file.read()
+        # TODO(BECATRUE): The code will be modifed in #37.
     else:
         pass
     expid = {
--- a/main.py
+++ b/main.py
@@ -2,7 +2,9 @@
 
 import json
 import logging
+import os
 import posixpath
+import shutil
 import time
 from contextlib import asynccontextmanager
 from datetime import datetime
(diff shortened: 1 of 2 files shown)


main.py
```diff
@@ -95,26 +97,14 @@ async def get_experiment_info(file: str) -> Any:
...
 @app.get("/experiment/submit/")
-async def submit_experiment(
+async def submit_experiment(  # pylint: disable=too-many-arguments
     file: str,
     args: str = "{}",
     pipeline: str = "main",
     priority: int = 0,
-    timed: Optional[str] = None
+    timed: Optional[str] = None,
+    visualize: Optional[bool] = False
```
kangz12345: Why should it be `Optional`?

main.py
```diff
@@ -126,10 +116,19 @@ async def submit_experiment(
...
     Returns:
         The run identifier, an integer which is incremented at each experiment submission.
     """
+    if visualize:
+        experiment_path = posixpath.join(configs["master_path"], configs["repository_path"], file)
+        with open(experiment_path, encoding="utf-8") as experiment_file:
+            _code = experiment_file.read()
+        # TODO(BECATRUE): The code will be modifed in #37.
+    else:
+        pass
```
kangz12345: Is this necessary? BTW why doesn't pylint complain about this?

kangz12345: Is this necessary? BTW why doesn't pylint complain about this?

kangz12345 (changes requested):

main.py
BECATRUE: Umm.. it is just for convenience! I intended the default options is `False`.

main.py
BECATRUE: This is necessary later becasue I will set the value of `file` in `expid` as a variable, and assign the appropriate path if the visualize option is `True` or not.

BECATRUE: > BTW why doesn't pylint complain about this?

I think if there is no `pass`, it will cause a syntax error, so in this case, Pylint seems to allow it.

main.py
kangz12345: `Optional[bool]` is `bool` or `None`. If you don't need `None`, you should set it as `bool`.

main.py
kangz12345: Okay then please add a TODO comment, since `else: pass` itself is a meaningless statement.

kangz12345 (changes requested):

main.py
BECATRUE: 😮 I was completely deluded! I will remove it.

kangz12345 (approved):
LGTM

BECATRUE:
Should I wait one more approval? @kangz12345

kangz12345:
> Should I wait one more approval? @kangz12345

Yes, I told @giwon2004 and he will review this until tomorrow.

main.py
```diff
@@ -126,10 +116,20 @@ async def submit_experiment(
         priority: Higher value means sooner scheduling.
         timed: The due date for the experiment in ISO format.
           None for no due date.
+        visualize: If True, the experiment file is modified for visualization.
+          The original file and vcd file are saved in the visualize path set in config.json.
     
     Returns:
         The run identifier, an integer which is incremented at each experiment submission.
     """
+    if visualize:
```
giwon2004, line 125: Why does this `if visualize:` appear twice?

Can it be unified with the next `if visualize:` block? Are there (or will be there) any prerequisites before defining `expid`?

giwon2004 (commented):
I have one simple question!

main.py
BECATRUE: Oh, it is a good question!

In the later PR, I will assign `expid` differently according to if `visualize` is `True` or `False`.
For this, I made `if visuzlie:` block before and after `remote.submit()`.

giwon2004: Hmm, I do not know how complex it is, but if the blocks are short enough, I would rather use:
```
if visualize:
    A
    B
    C
else:
    B
```
instead of:
```
if visualize:
    A
B
if visualize:
    C
```
Is the `B` part so complex that it must be shared by two situations?

giwon2004: I guess this can be handled in future PR. I will approve this merge.

giwon2004 (approved):

main.py
kangz12345: In this case, I am not sure that your suggestion works because there is an `else:` for the first `if:`.
In addition, since this project is not that performance-critical, we prefer readability to performance unless it is way too inefficient.
I think `if visualize:` is not expensive in such a sense!

PR #41, opened 2023-08-02
Make an experiment fetcher

### LINKED ISSUE #29: Experiment queue fetcher for display in `iquip/apps/scheduler.py`
From `artiq_master`, the experiment queue is returned when `$ artiq_client show schedule`.
Here, it is again implemented in `main.py` to fetch this information and send it to `scheduler.py` to display the queued experiments.

giwon2004:
The function `get_experiment_queue()` returns a dictionary like:
```
{'167': {'pipeline': 'main', 'expid': {'log_level': 30, 'class_name': None, 'arguments': {'user': 'QuIQCL', 'time': 1.0, 'save': False, 'color': 'r'}, 'file': 'DIRECTORY-PATH'}, 'priority': 1, 'due_date': None, 'flush': False, 'status': 'analyzing', 'repo_msg': None}, 
 '168': {'pipeline': 'main', 'expid': {'log_level': 30, 'class_name': None, 'arguments': {'user': 'QuIQCL', 'time': 1.0, 'save': False, 'color': 'r'}, 'file': 'DIRECTORY-PATH'}, 'priority': 1, 'due_date': None, 'flush': False, 'status': 'pending', 'repo_msg': None}, 
 '169': {'pipeline': 'main', 'expid': {'log_level': 30, 'class_name': None, 'arguments': {'user': 'QuIQCL', 'time': 1.0, 'save': False, 'color': 'r'}, 'file': 'DIRECTORY-PATH'}, 'priority': 1, 'due_date': None, 'flush': False, 'status': 'pending', 'repo_msg': None}, 
 '170': {'pipeline': 'main', 'expid': {'log_level': 30, 'class_name': None, 'arguments': {'user': 'QuIQCL', 'time': 1.0, 'save': False, 'color': 'r'}, 'file': 'DIRECTORY-PATH'}, 'priority': 1, 'due_date': None, 'flush': False, 'status': 'pending', 'repo_msg': None}}
```

which is the dictionary of scheduled experiments, whose keys are the`rid`value.

This dictionary is fetched from `scheduler.py` (introduced in issue [#72](https://github.com/snu-quiqcl/iquip/issues/72) in`iquip`and pull request [#91](https://github.com/snu-quiqcl/iquip/pull/91)) to display these experiments.

This closes #29.

### DIFF
--- a/main.py
+++ b/main.py
@@ -95,6 +95,19 @@ async def get_experiment_info(file: str) -> Any:
     return remote.examine(file)
 
 
+@app.get("/experiment/queue/", response_model=Dict)
+async def get_experiment_queue() -> Dict:
+    """Gets the list of queued experiment returns it.
+
+    Returns:
+        A dictionary of queued experiments with rid as their keys.
+        It includes the running experiment with different "status" value.
+    """
+    remote = get_client("master_schedule")
+    current_experiment = remote.get_status()
+    return current_experiment
+
+
 @app.get("/experiment/code/")
 async def get_experiment_code(file: str) -> str:
     """Gets code of the given experiment file and returns it.
--- a/main.py
+++ b/main.py
@@ -95,8 +95,8 @@ async def get_experiment_info(file: str) -> Any:
     remote = get_client("master_experiment_db")
     return remote.examine(file)
 
-
-current_queue = previous_queue = dict()
+previous_queue = {}
+current_queue = {}
 @app.get("/experiment/queue/", response_model=Dict[int, Dict[str, Any]])
 async def get_experiment_queue() -> Dict[int, Dict[str, Any]]:
     """Gets the list of queued experiment and returns it.
@@ -107,12 +107,13 @@ async def get_experiment_queue() -> Dict[int, Dict[str, Any]]:
           "priority", "status", "flush", "pipeline", "expid", and etc.
         It includes the running experiment with different "status" value.
     """
-    global current_queue, previous_queue
     remote = get_client("master_schedule")
     while current_queue == previous_queue:
         await asyncio.sleep(0)
-        current_queue = remote.get_status()
-    previous_queue = current_queue
+        current_queue.clear()
+        current_queue.update(remote.get_status())
+    previous_queue.clear()
+    previous_queue.update(current_queue)
     return current_queue
 
 
--- a/main.py
+++ b/main.py
@@ -95,6 +95,7 @@ async def get_experiment_info(file: str) -> Any:
     remote = get_client("master_experiment_db")
     return remote.examine(file)
 
+
 previous_queue = {}
 current_queue = {}
 @app.get("/experiment/queue/", response_model=Dict[int, Dict[str, Any]])
--- a/main.py
+++ b/main.py
@@ -4,6 +4,7 @@ import json
 import logging
 import posixpath
 import time
+import asyncio
 from contextlib import asynccontextmanager
 from datetime import datetime
 from typing import Any, Dict, List, Optional
@@ -95,8 +96,9 @@ async def get_experiment_info(file: str) -> Any:
     return remote.examine(file)
 
 
-@app.get("/experiment/queue/", response_model=Dict[int, Dict])
-async def get_experiment_queue() -> Dict[int, Dict]:
+current_queue = previous_queue = dict()
+@app.get("/experiment/queue/", response_model=Dict[int, Dict[str, Any]])
+async def get_experiment_queue() -> Dict[int, Dict[str, Any]]:
     """Gets the list of queued experiment and returns it.
 
     Returns:
@@ -105,8 +107,13 @@ async def get_experiment_queue() -> Dict[int, Dict]:
           "priority", "status", "flush", "pipeline", "expid", and etc.
         It includes the running experiment with different "status" value.
     """
+    global current_queue, previous_queue
     remote = get_client("master_schedule")
-    return remote.get_status()
+    while current_queue == previous_queue:
+        await asyncio.sleep(0)
+        current_queue = remote.get_status()
+    previous_queue = current_queue
+    return current_queue
 
 
 @app.get("/experiment/code/")
--- a/main.py
+++ b/main.py
@@ -95,17 +95,18 @@ async def get_experiment_info(file: str) -> Any:
     return remote.examine(file)
 
 
-@app.get("/experiment/queue/", response_model=Dict)
-async def get_experiment_queue() -> Dict:
-    """Gets the list of queued experiment returns it.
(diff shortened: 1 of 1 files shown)


BECATRUE:
You can use several keywords, i.e. `close` and `resolve` to close a specific issue when the PR is merged.
Please see [Linking a PR to an issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#about-linked-issues-and-pull-requests).

main.py
```diff
@@ -95,6 +95,19 @@ async def get_experiment_info(file: str) -> Any:
     return remote.examine(file)
 
 
+@app.get("/experiment/queue/", response_model=Dict)
```
BECATRUE: Please write down the type annotation in more detail.

main.py
```diff
@@ -95,6 +95,19 @@ async def get_experiment_info(file: str) -> Any:
...
+async def get_experiment_queue() -> Dict:
+    """Gets the list of queued experiment returns it.
```
BECATRUE: Did you intend "... experiments and returns ..."?

main.py
```diff
@@ -95,6 +95,19 @@ async def get_experiment_info(file: str) -> Any:
...
+
+    Returns:
+        A dictionary of queued experiments with rid as their keys.
```
BECATRUE, line 109: Please write it down in more detail.
For example, what is each value and if the value is a dictionary, then what are its keys and values.

main.py
```diff
@@ -95,6 +95,19 @@ async def get_experiment_info(file: str) -> Any:
...
+        It includes the running experiment with different "status" value.
+    """
+    remote = get_client("master_schedule")
+    current_experiment = remote.get_status()
```
BECATRUE: Wow, how can you find this method😮

main.py
```diff
@@ -95,6 +95,19 @@ async def get_experiment_info(file: str) -> Any:
...
+    return current_experiment
```
BECATRUE: Please return directly, not assign it to an unnecessary variable.

BECATRUE (changes requested):
How does the IQUIP scheduler update its status?
Does it request to the proxy server periodically?

main.py
giwon2004: By backtracking..? 😂

giwon2004:
> How does the IQUIP scheduler update its status? Does it request to the proxy server periodically?

Yes, that is what I have made so far, but it is subject to change: see [discussion #90](https://github.com/snu-quiqcl/iquip/discussions/90) in iquip.

main.py
kangz12345: E.g., `Dict[str, Any]` (please use the exact type, this is just an example).

giwon2004: Is it resolved now?

BECATRUE: Umm.. how about `Dict[int, Dict[str, Any]]`?

giwon2004: Okay, I will fix it that way.

giwon2004:
Please check [Discussion #94](https://github.com/snu-quiqcl/iquip/discussions/94) in iquip.

main.py
```diff
@@ -95,6 +96,28 @@ async def get_experiment_info(file: str) -> Any:
     return remote.examine(file)
 
 
+previous_queue = {}
+current_queue = {}
+@app.get("/experiment/queue/", response_model=Dict[int, Dict[str, Any]])
```
BECATRUE: Please add two empty lines after defining `current_queue`.

BECATRUE: Also, as I know, the `response_model` is only necessary when it returns a Pydantic model.
Please check it!

main.py
```diff
@@ -95,6 +96,28 @@ async def get_experiment_info(file: str) -> Any:
...
+async def get_experiment_queue() -> Dict[int, Dict[str, Any]]:
+    """Gets the list of queued experiment and returns it.
+
+    Returns:
+        A dictionary of queued experiments with rid as their keys.
+        The value of each corresponding rid is a dictionary with several information:
+          "priority", "status", "flush", "pipeline", "expid", and etc.
```
BECATRUE: It is a minor suggestion, but please remove "and".
The "etc" includes the meaning of "and".

BECATRUE: There is the `timed` (or due date) key, right?

I think we need the information, so it looks better to include this key in the above docstring!
And it is okay to remove "fluse" as we probably won't use it.

BECATRUE (changes requested):

BECATRUE (dismissed):
LGTM!

kangz12345 (approved):
OK for now but we still have to consider:
1. Efficiency: does `remove.get_status()` return immediately or not? If it does, the RPC happens too frequently.
2. Timeout: I think it is more stable if a timeout is set. For example, when the queue is never changed, then it will not respond back, and the socket connection might disappear.

main.py
```diff
@@ -95,6 +96,30 @@ async def get_experiment_info(file: str) -> Any:
     return remote.examine(file)
 
 
+previous_queue = {}
+current_queue = {}
```
kangz12345: Do we need both dictionaries? I think 1 is enough to store the previous state. The current state can be handled locally I guess.

kangz12345 (changes requested):

kangz12345:
I don't know why it has conflicts but please resolve it as well!

giwon2004:
> OK for now but we still have to consider:
> 
> 1. Efficiency: does `remote.get_status()` return immediately or not? If it does, the RPC happens too frequently.
> 2. Timeout: I think it is more stable if a timeout is set. For example, when the queue is never changed, then it will not respond back, and the socket connection might disappear.

1. It does, but status changes like `run_done` to `analyzing` is almost at that frequency too. Though I think this calls too frequently. (That is why we use `Subscriber`, because we do not have to request again and again to track changes.)
2. Okay, I agree. I will do that in my `scheduler.py` branch.

giwon2004:
I will resolve conflicts after PR #46 is merged.

kangz12345:
> 1. It does, but status changes like `run_done` to `analyzing` is almost at that frequency too. Though I think this calls too frequently. (That is why we use `Subscriber`, because we do not have to request again and again to track changes.)

Exactly! I wonder why you don't use `Subscriber` here.
However, for now, I think we need to implement more features instead of optimizing the performance. Please move on to your next goal.

kangz12345 (dismissed):
LGTM

kangz12345:
Please resolve the conflict. I think `get_experiment_code()` should be removed again.

giwon2004:
Done it!

kangz12345 (approved):

BECATRUE (approved):
LGTM!

PR #46, opened 2023-08-06
A `scheduler` replica

### LINKED ISSUE #44: Make a replica of a `Scheduler` class that delivers all function call request to the artiq server
As mentioned in [Issue#96](https://github.com/snu-quiqcl/iquip/issues/96) of `iquip`, I need a port to send the modify request of the queued experiment to the _real_ scheduler in the server.

In `main.py`, I will make a POST (maybe GET if it needs return value) request path that receives a function call request of functions in [`artiq/artiq/master/scheduler.py`](https://github.com/m-labs/artiq/blob/master/artiq/master/scheduler.py#L400) and calls the functions in scheduler that is connected by `remote = get_client("master_schedule")`.

giwon2004:
It is a simple replica of a `scheduler` in [`artiq/master/scheduler.py`](https://github.com/m-labs/artiq/blob/master/artiq/master/scheduler.py#L400) to receive signals to delete experiments and request termination of experiments from our GUI.

There are two functions implemented:
`delete_experiment(rid)`: Kills the run with the specified RID.
`request_termination_of_experiment(rid)`: Requests graceful termination of the run with the specified RID.

Both are already defined as a method of `Scheduler` instance in [`artiq/master/scheduler.py`](https://github.com/m-labs/artiq/blob/master/artiq/master/scheduler.py#L400) and above two calls each one.

This closes #44.

### DIFF
--- a/main.py
+++ b/main.py
@@ -109,22 +109,22 @@ async def get_experiment_code(file: str) -> str:
 
 
 @app.post("/experiment/delete/")
-async def delete_experiment(rid):
+async def delete_experiment(rid: int):
     """Kills the run with the specified RID.
 
     Args:
-        rid: rid value of the target experiment.
+        rid: The run identifier value of the target experiment.
     """
     remote = get_client("master_schedule")
     remote.delete(rid)
 
 
 @app.post("/experiment/request_termination/")
-async def request_termination_of_experiment(rid):
+async def request_termination_of_experiment(rid: int):
     """Requests graceful termination of the run with the specified RID.
 
     Args:
-        rid: rid value of the target experiment.
+        rid: The run identifier value of the target experiment.
     """
     remote = get_client("master_schedule")
     remote.request_termination(rid)
--- a/main.py
+++ b/main.py
@@ -110,14 +110,22 @@ async def get_experiment_code(file: str) -> str:
 
 @app.post("/experiment/delete/")
 async def delete_experiment(rid):
-    """Kills the run with the specified RID."""
+    """Kills the run with the specified RID.
+
+    Args:
+        rid: rid value of the target experiment.
+    """
     remote = get_client("master_schedule")
     remote.delete(rid)
 
 
 @app.post("/experiment/request_termination/")
-async def delete_experiment(rid):
-    """Requests graceful termination of the run with the specified RID."""
+async def request_termination_of_experiment(rid):
+    """Requests graceful termination of the run with the specified RID.
+
+    Args:
+        rid: rid value of the target experiment.
+    """
     remote = get_client("master_schedule")
     remote.request_termination(rid)
 
--- a/main.py
+++ b/main.py
@@ -97,19 +97,6 @@ async def get_experiment_info(file: str) -> Any:
     return remote.examine(file)
 
 
-@app.get("/experiment/code/")
-async def get_experiment_code(file: str) -> str:
-    """Gets code of the given experiment file and returns it.
-    
-    Args:
-        file: The path of the experiment file.
-    """
-    full_path = posixpath.join(configs["master_path"], configs["repository_path"], file)
-    with open(full_path, encoding="utf-8") as experiment_file:
-        code = experiment_file.read()
-    return code
-
-
 @app.post("/experiment/delete/")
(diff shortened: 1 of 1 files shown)


main.py
```diff
@@ -108,6 +108,28 @@ async def get_experiment_code(file: str) -> str:
     return code
 
 
+@app.post("/experiment/delete/")
+async def delete_experiment(rid):
```
BECATRUE: Please note the type hint.

main.py
```diff
@@ -108,6 +108,28 @@ async def get_experiment_code(file: str) -> str:
...
+    """Kills the run with the specified RID.
+
+    Args:
+        rid: rid value of the target experiment.
```
BECATRUE: How about "The run identifier of the ..."?

main.py
```diff
@@ -108,6 +108,28 @@ async def get_experiment_code(file: str) -> str:
...
+    """
+    remote = get_client("master_schedule")
+    remote.delete(rid)
+
+
+@app.post("/experiment/request_termination/")
```
BECATRUE: How about "/experiment/terminate/"?
It looks too long.

BECATRUE: Umm.. Why do you need this? I think `remote.delete(rid)` is enough..

Is there any reason?

BECATRUE (changes requested):

main.py
giwon2004: The original `scheduler` supports these two different functions, and I guess there is a difference like one between Alt+F4 and quitting by clicking the X button.

BECATRUE: What is difference betwwen `delete()` and `request_termination()`?

giwon2004: I do not know in detail, but I made a _replica_ so I thought I need to make them for future use.

BECATRUE: @kangz12345 How about you?

BECATRUE:
Also, I think `get` is more appropriate even if it doesn't need to return something, because it doesn't create or change the server.

main.py
kangz12345: See also: [source code](https://m-labs.hk/artiq/manual/_modules/artiq/master/scheduler.html#Scheduler.delete)

`request_termination()` will do some proper works before deleting an experiment.
I think both are required (`delete()` could be like force-quit).

kangz12345:
[RFC9110](https://www.rfc-editor.org/rfc/rfc9110#name-method-definitions) would be helpful to decide which method to use. Moreover, it would be a good practice for us to read the whole document.

giwon2004:
I guess `post` is enough. It is better for security too.

BECATRUE (dismissed):
LGTM!

Did you check if both requests work well?

kangz12345 (dismissed):

BECATRUE:
@giwon2004 Oh, because of merging #36 to `develop` branch, the conflicts occurs.
Please fix this.

kangz12345:
You can just click the "Resolve conflicts" button and remove "====" or "<<<<" things to make the final version.

giwon2004:
> You can just click the "Resolve conflicts" button and remove "====" or "<<<<" things to make the final version.

I clicked **Mark as resolved**, but the conflict did not resolve.
So I clicked **Merge commit**. Is this what you meant to do?

BECATRUE:
> > You can just click the "Resolve conflicts" button and remove "====" or "<<<<" things to make the final version.
> 
> I clicked **Mark as resolved**, but the conflict did not resolve. So I clicked **Merge commit**. Is this what you meant to do?

Probably you are right! Then, I will leave reviews again!

main.py
```diff
@@ -97,6 +97,41 @@ async def get_experiment_info(file: str) -> Any:
     return remote.examine(file)
 
 
+@app.get("/experiment/code/")
+async def get_experiment_code(file: str) -> str:
```
BECATRUE: Umm.. Why was this code created?

If you handled this during resolving the conflicts, then please remove this function!

BECATRUE (changes requested):

BECATRUE (approved):
LGTM!

kangz12345 (approved):

Issue #38, opened 2023-08-01
Send the result directory structure


BECATRUE:
### Feature you want to implement
- Similar to `/ls/`, send the result directory structure.
- Once receiving a request from a client, re-organize files in the result directory.

### How the feature is implemented
1. Read `rid.json` in the result directory.
It contains the most recently fetched RID.
2. Navigate to all RID directories after the RID above.
3. In each directory, read the metafile and copy the corresponding h5 result file into this directory.
4. Return the RID list.

PR #50, opened 2023-08-13
Modify an experiment to make logs and vcd file

### LINKED ISSUE #37: Modify an experiment to make logs and vcd file
- Add logs to track each line.
- Fetch one's RID.
- Generate the vcd file.
- Submit the modified experiment.

### LINKED ISSUE #42: Save the submission information
- If the visualize option is set, save the submission information in the corresponding directory, i.e., `${master_path}/{visualize_path}/{RID}/`.

- Save all information to `metadata.json`.
- The information is as follows:
  - Submission time

BECATRUE:
This PR will close #37 and close #42.

Fisrt of all, I apologize for the long and convoluted code😅

I implemented the following two main features:
1. Modify the experiment to make logs and vcd file using [`ast`](https://docs.python.org/3/library/ast.html) library.
2. Save the submission time into the metadata file.

# Overall procedure
1. Read the original file and modify it by `modify_experiment_code()`.
2. Save the modified experiment in `{visualize_path}/experiment_{submission_time}.py`.
This file will be moved in `{visualize_path}/{rid}/modified_experiment.py` after fetching the result directory.
#38
3. Submit the modified experiment with the corresponding experiment path.
The vcd file is created in `{visualize_path}/{rid}/rtio.vcd`.
4. Make the visualize result directory with the path `{visualize_path}/{rid}/`.
5. Save the metadata in `{visualize_path}/{rid}/metadata.json`.

# Modify the experiment
This uses `ast` library basically.

To leave an rtio log, I used [`rtio_log`](https://m-labs.hk/artiq/manual/getting_started_core.html#rtio-analyzer-example).
All logs are created just before each line in `run()`, and contain the line number information.

To make a vcd file, I used [`artiq_coreanalyzer`](https://m-labs.hk/artiq/manual/utilities.html#core-device-rtio-analyzer-tool).
This feature is written in `write_vcd()`.

For example, the original code is as follows.
```python

class TTLWave(EnvExperiment):
    """Simple TTL pulses"""

    def build(self):
        self.setattr_device("core")
        self.setattr_device("ttl4")
        self.dds0 = self.get_device("urukul0_ch0")
        self.setattr_argument("count", NumberValue(default=3, min=1, max=1000, ndecimals=0, scale=1, step=1, type="int"))
        self.setattr_argument("time", NumberValue(default=10*10**-3, min=1*10**-3, max=1, unit="ms"))

    @kernel
    def run(self):
        self.core.reset()
        for i in range(self.count):
            self.ttl4.on()
            delay(self.time)
            self.ttl4.off()
            delay(self.time)
```

The modified code is as follows:
```python
import subprocess
import posixpath

class TTLWave(EnvExperiment):
    """Simple TTL pulses"""

    def build(self):
        self.setattr_device('core')
        self.setattr_device('ttl4')
        self.dds0 = self.get_device('urukul0_ch0')
        self.setattr_argument('count', NumberValue(default=3, min=1, max=1000, ndecimals=0, scale=1, step=1, type='int'))
        self.setattr_argument('time', NumberValue(default=10 * 10 ** (-3), min=1 * 10 ** (-3), max=1, unit='ms'))

    @kernel
    def run(self):
        rtio_log('line', 15)
        self.core.reset()
        rtio_log('line', 16)
        for i in range(self.count):
            rtio_log('line', 17)
            self.ttl4.on()
            rtio_log('line', 18)
            delay(self.time)
            rtio_log('line', 19)
            self.ttl4.off()
            rtio_log('line', 20)
            delay(self.time)
        self.write_vcd()

    def write_vcd(self):
        result = subprocess.run('curl http://127.0.0.1:8000/experiment/running/', capture_output=True, shell=True, text=True)
        rid = int(result.stdout)
        device_db_path = posixpath.join('/home/rabi/artiq/', 'device_db.py')
        vcd_path = posixpath.join('/home/rabi/artiq/', 'visualize/', f'{rid}/rtio.vcd')
        subprocess.run(f'artiq_coreanalyzer --device-db {device_db_path} -w {vcd_path}', capture_output=True, shell=True)
```

# Save the metadata file
After each submission, the submission time is saved in `{visualize_path}/{rid}/metadata.json`.

An example file is as follows:
```json
{"submission_time": "2023-08-13T14:51:51"}
```

The information contained is small, but it can be expanded later, so I saved it as json.

# Result
## Visualize directory structure
The visualize path is set to `visualize/` in `config.json` and the directory is created in the artiq-master path.

The directory structure is as follows.

## VCD file
According to the above experiment file, it will make three pulses and leave the logs with the line number 18 (1st delay) and 20 (2nd delay).

# Note
The modified experiment __CANNOT__ be performed in an environment using an ARTIQ simulator.
To run `rtio_log()`, you should run this with the actual ARTIQ devices.

### DIFF
--- a/main.py
+++ b/main.py
@@ -1,5 +1,6 @@
 """Proxy server to communicate a client to ARTIQ."""
 
+import ast
 import json
 import logging
 import os
@@ -144,9 +145,19 @@ async def request_termination_of_experiment(rid: int):
     remote.request_termination(rid)
 
 
+def modify_experiment_code(code: str, experiment_cls_name: str):
+    code_ast = ast.parse(code)
+    experiment_cls_ast = next(
+        stmt for stmt in code_ast.body
+        if isinstance(stmt, ast.ClassDef) and stmt.name == experiment_cls_name
+    )
+    return ast.unparse(code_ast)
+
+
 @app.get("/experiment/submit/")
 async def submit_experiment(  # pylint: disable=too-many-arguments
     file: str,
+    cls: str,
     args: str = "{}",
     pipeline: str = "main",
     priority: int = 0,
@@ -157,6 +168,7 @@ async def submit_experiment(  # pylint: disable=too-many-arguments
     
     Args:
         file: The path of the experiment file.
+        cls: The class name of the experiment.
         args: The arguments to submit which must be a JSON string of a dictionary.
           Each key is an argument name and its value is the value of the argument.
         pipeline: The pipeline to run the experiment in.
@@ -173,7 +185,8 @@ async def submit_experiment(  # pylint: disable=too-many-arguments
         experiment_path = posixpath.join(configs["master_path"], configs["repository_path"], file)
         with open(experiment_path, encoding="utf-8") as experiment_file:
             code = experiment_file.read()
-        modified_code = modify_experiment_code(code)
+        modified_code = modify_experiment_code(code, cls)
+        print(modified_code)
     else:
         # TODO(BECATRUE): The exact experiment path will be assigned in #37.
         pass
--- a/main.py
+++ b/main.py
@@ -201,7 +201,7 @@ import posixpath
     # make logs to track each line
     run_func_stmt.body = add_tracking_line(run_func_stmt.body)
     # call write_vcd()
-    write_vcd_call_stmt = ast.parse("write_vcd()").body
+    write_vcd_call_stmt = ast.parse("self.write_vcd()").body
     run_func_stmt.body.append(write_vcd_call_stmt)
     # define write_vcd()
     write_vcd_func_code = f"""
--- a/main.py
+++ b/main.py
@@ -160,6 +160,9 @@ def add_tracking_line(stmt_list: List[ast.stmt]) -> List[ast.stmt]:
 
 def modify_experiment_code(code: str, experiment_cls_name: str):
     code_mod = ast.parse(code)
+    # import subprocess module
+    import_subprocess_stmt = ast.parse("import subprocess").body
+    code_mod.body.insert(0, import_subprocess_stmt)
     experiment_cls_stmt = next(
         stmt for stmt in code_mod.body
         if isinstance(stmt, ast.ClassDef) and stmt.name == experiment_cls_name
--- a/main.py
+++ b/main.py
@@ -151,6 +151,10 @@ def modify_experiment_code(code: str, experiment_cls_name: str):
         stmt for stmt in code_ast.body
         if isinstance(stmt, ast.ClassDef) and stmt.name == experiment_cls_name
     )
+    run_func_ast = next(
+        stmt for stmt in experiment_cls_ast.body
+        if isinstance(stmt, ast.FunctionDef) and stmt.name == "run"
+    )
     return ast.unparse(code_ast)
 
 
--- a/main.py
+++ b/main.py
@@ -251,24 +251,21 @@ async def submit_experiment(  # pylint: disable=too-many-arguments
         The run identifier, an integer which is incremented at each experiment submission.
     """
     if visualize:
-        visualize_dir_path = posixpath.join(
-            configs["master_path"],
-            configs["visualize_path"],
-            f"{rid}/"
-        )
-        os.makedirs(visualize_dir_path)
+        submission_time = datetime.now().isoformat(timespec="seconds")
         experiment_path = posixpath.join(configs["master_path"], configs["repository_path"], file)
-        # copy the original experiment file
-        copied_experiment_path = posixpath.join(visualize_dir_path, "experiment.py")
-        shutil.copyfile(experiment_path, copied_experiment_path)
+        visualize_dir_path = posixpath.join(configs["master_path"], configs["visualize_path"])
+        modified_experiment_path = posixpath.join(
+            visualize_dir_path,
+            f"experiment_{submission_time}.py"
+        )
         # modify the experiment code
         with open(experiment_path, encoding="utf-8") as experiment_file:
             code = experiment_file.read()
         modified_code = modify_experiment_code(code, cls)
         # save the modified experiment code
-        modified_experiment_path = posixpath.join(visualize_dir_path, "modified_experiment.py")
         with open(modified_experiment_path, "w", encoding="utf-8") as modified_experiment_file:
             modified_experiment_file.write(modified_code)
+        submission_file_path = modified_experiment_path
     else:
         # TODO(BECATRUE): The exact experiment path will be assigned in #37.
         pass
@@ -283,15 +280,16 @@ async def submit_experiment(  # pylint: disable=too-many-arguments
     remote = get_client("master_schedule")
     rid = remote.submit(pipeline, expid, priority, due_date, False)
     if visualize:
+        rid_dir_path = posixpath.join(visualize_dir_path, f"{rid}/")
+        os.makedirs(rid_dir_path)
+        # copy the original experiment file
+        copied_experiment_path = posixpath.join(rid_dir_path, "experiment.py")
+        shutil.copyfile(experiment_path, copied_experiment_path)
+        # save the metadata
         metadata = {
-            "file": experiment_path,
-            "args": args_dict,
-            "pipeline": pipeline,
-            "priority": priority,
-            "timed": timed,
-            "submission_time": datetime.now().isoformat(timespec="seconds")
+            "submission_time": submission_time
         }
-        metadata_path = posixpath.join(visualize_dir_path, "metadata.json")
+        metadata_path = posixpath.join(rid_dir_path, "metadata.json")
         with open(metadata_path, "w", encoding="utf-8") as metadata_file:
             json.dump(metadata, metadata_file)
     return rid
--- a/main.py
+++ b/main.py
@@ -184,16 +184,16 @@ def modify_experiment_code(code: str, experiment_cls_name: str) -> str:
     Returns:
         The modified experiment code.
     """
-    code_mod = ast.parse(code)
+    mod = ast.parse(code)
     # import modules
     import_code = """
 import subprocess
 import posixpath
     """
     import_stmt = ast.parse(import_code).body
-    code_mod.body = import_stmt + code_mod.body
+    mod.body = import_stmt + mod.body
     experiment_cls_stmt = next(
-        stmt for stmt in code_mod.body
+        stmt for stmt in mod.body
         if isinstance(stmt, ast.ClassDef) and stmt.name == experiment_cls_name
     )
     run_func_stmt = next(
@@ -225,7 +225,7 @@ def write_vcd(self):
     """
     write_vcd_func_stmt = ast.parse(write_vcd_func_code).body
     experiment_cls_stmt.body.append(write_vcd_func_stmt)
-    return ast.unparse(code_mod)
+    return ast.unparse(mod)
 
 
 @app.get("/experiment/submit/")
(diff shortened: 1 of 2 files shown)


BECATRUE:
I will leave more reviews tonight😅

giwon2004:
Where can I find your 'modified code' in your first merge comment?

giwon2004:
Can you tell me about `rtio_log` function in detail? In the document you linked, it used.
```rtio_log("ttl0", "i", i)```
which is quite different from your way of use.

Thank you in advance.

File: main.py [hunk 1]
```diff
@@ -144,14 +145,96 @@ async def request_termination_of_experiment(rid: int):
...
+def add_tracking_line(stmt_list: List[ast.stmt]) -> List[ast.stmt]:
+    """Modifies the given statements list to leave rtio logs that track each line.
+
+    Adds logs containing the corresponding line number before each line, recursively.
+
+    Args:
+        stmt_list: The list of ast statements.
+
+    Returns:
+        The list of modified ast statements.
```
kangz12345: This explanation might be interpreted as it mutates the given list. Please clarify that it returns a _new_ list, leaving the given list intact.

main.py
```diff
@@ -144,14 +145,96 @@ async def request_termination_of_experiment(rid: int):
...
+    """
+    modified_stmt_list = []
+    for stmt in stmt_list:
+        tracking_line_code = f"rtio_log('line', {stmt.lineno})"
+        tracking_line_stmt = ast.parse(tracking_line_code).body
+        modified_stmt_list.extend(tracking_line_stmt)
+        for attr in ("body", "handlers", "orelse", "finalbody"):
+            if hasattr(stmt, attr):
+                setattr(stmt, attr, add_tracking_line(getattr(stmt, attr)))
```
kangz12345, line 168: 👍

main.py
```diff
@@ -144,14 +145,96 @@ async def request_termination_of_experiment(rid: int):
...
+        2. Make the vcd file into the corresponding path.
+
+    Args:
+        code: The experiment code.
+        experiment_cls_name: The class name of the experiment.
+
+    Returns:
+        The modified experiment code.
+    """
+    code_mod = ast.parse(code)
```
kangz12345: The uses of "code" are quite confusing here.
You used `code_` to show that this is a `mod` parsed from the `code`.
However, in the rest part of this function, `_code` is used for string code representation.
How about just naming it `mod` instead of `code_mod`?

main.py
```diff
@@ -144,14 +145,96 @@ async def request_termination_of_experiment(rid: int):
...
+    import_stmt = ast.parse(import_code).body
+    code_mod.body = import_stmt + code_mod.body
+    experiment_cls_stmt = next(
+        stmt for stmt in code_mod.body
+        if isinstance(stmt, ast.ClassDef) and stmt.name == experiment_cls_name
+    )
+    run_func_stmt = next(
+        stmt for stmt in experiment_cls_stmt.body
+        if isinstance(stmt, ast.FunctionDef) and stmt.name == "run"
+    )
```
kangz12345: Please make a note in the docstring that the code must contain the experiment class and `run()` method.
Or, you may handle those exceptions if you want to.

main.py
```diff
@@ -165,36 +248,53 @@ async def submit_experiment(  # pylint: disable=too-many-arguments
...
+        # save the modified experiment code
+        with open(modified_experiment_path, "w", encoding="utf-8") as modified_experiment_file:
+            modified_experiment_file.write(modified_code)
+        submission_file_path = modified_experiment_path
     else:
-        # TODO(BECATRUE): The exact experiment path will be assigned in #37.
-        pass
+        submission_file_path = posixpath.join(configs["repository_path"], file)
+    args_dict = json.loads(args)
     expid = {
```
kangz12345, line 271: What was `expid`? Please rename it with full words.

main.py
```diff
@@ -144,14 +145,96 @@ async def request_termination_of_experiment(rid: int):
...
+    # define write_vcd()
+    write_vcd_func_code = f"""
+def write_vcd(self):
+    result = subprocess.run("curl http://127.0.0.1:8000/experiment/running/",
+                            capture_output=True, shell=True, text=True)
+    rid = int(result.stdout)
+    device_db_path = posixpath.join(
+        "{configs["master_path"]}",
+        "device_db.py"
+    )
```
kangz12345: You can `join()` it in advance and give the result path as a constant string.
Then we don't have to import `posixpath` in the experiment file, do we?

kangz12345 (changes requested):

main.py
giwon2004: As far as I know, the name `expid` came from the original artiq server.

main.py
BECATRUE: Oh, you're right! I will fix it.

main.py
BECATRUE: Yes! As Giwon said, `expid` came from [`artiq_client.py`](https://github.com/m-labs/artiq/blob/master/artiq/frontend/artiq_client.py#L148).
Though it is an abbreviation and looks little vague, it looks better to use the original expression.

BECATRUE:
> Where can I find your 'modified code' in your first merge comment?

Could you explain more details?

BECATRUE:
> Can you tell me about rtio_log function in detail?

As I presented yesterday, `rtio_log(label, *args)` leaves a string log based on ASCII code with the label.
For example, `rtio_log("label1", "i", 1)` makes a log, "i 1" with the label "label1".

Also, as I tested, there is no restriction for the log length.

main.py
BECATRUE: Oh, it would be easier to understand!

main.py
BECATRUE: It will rarely happen if these conditions are not met, so the docsting is likely to be sufficient.

main.py
BECATRUE: I realize now that we need to import `posixpath` in order to define `vcd_path`.
It uses the RID so that we cannot define `vcd_path` in advance.

What do you think of it? @kangz12345

BECATRUE:
The unittest will be modified in the later issue.

But, we have some problem in terms of `ast.unparse()` method.
This is supported after Python 3.8, so currently Pylint for Python 3.8 makes an error.

How about supporting only after Python 3.8?

BECATRUE:
Also, I set the result path to `results/`, so the original result directory created by artiq-master and the RID directory are located in the same path as follows.

image

Is it okay?

giwon2004:
> How about supporting only after Python 3.8?

Check out [`astunparse`](https://astunparse.readthedocs.io/en/latest/); this might help.

giwon2004:
> > Where can I find your 'modified code' in your first merge comment?
> 
> Could you explain more details?

I meant this one.

image

BECATRUE:
> > How about supporting only after Python 3.8?
> 
> Check out [`astunparse`](https://astunparse.readthedocs.io/en/latest/); this might help.

Oh, thanks for the kind reference! Let's decide the specific solution to @kangz12345.

BECATRUE:
> > > Where can I find your 'modified code' in your first merge comment?
> > 
> > 
> > Could you explain more details?
> 
> I meant this one.
> 
> image

As I understand, you wonder the exact path of the modified one.

It is created in the result path with the name format, `experiment_{submission_time}.py`.
For example, there are two modified experiment files, `experiment_2023-08-14T{HH}:{MM}:{SS}.py` in the following image.

main.py
```diff
@@ -144,14 +146,99 @@ async def request_termination_of_experiment(rid: int):
...
+        The created list of ast statements.
+    """
+    modified_stmt_list = []
+    for stmt in stmt_list:
+        tracking_line_code = f"rtio_log('line', {stmt.lineno})"
+        tracking_line_stmt = ast.parse(tracking_line_code).body
+        modified_stmt_list.extend(tracking_line_stmt)
+        for attr in ("body", "handlers", "orelse", "finalbody"):
+            if hasattr(stmt, attr):
+                setattr(stmt, attr, add_tracking_line(getattr(stmt, attr)))
```
giwon2004, line 168: I do not understand, why doesn't this make an infinite loop?

BECATRUE: For example, it operates as follows:
```python
stmt.body = add_tracking_line(stmt.body)
```
In this case, the `stmt_list` of the callee is `stmt.body`.
Thus, it works well recursively.

Which part are you curious about?

giwon2004: Oh, I thought `stmt` is called again. I was confused. 😅

giwon2004 (approved):
I do not know well about `artiq_coreanalyzer` so I am not sure about it, but other than that, LGTM!

giwon2004:
Are these errors of `result_path` in the unit tests scheduled for another issue/PR?

main.py
kangz12345: If it rarely _happens_, it should be handled! Or it should be guaranteed that such exceptional cases never happen.

main.py
kangz12345: Even though RID cannot be determined in advance, you can still join the path without knowing it: e.g., `"dir1/dir2/{RID}/..."`.

kangz12345:
> > > How about supporting only after Python 3.8?
> > 
> > 
> > Check out [`astunparse`](https://astunparse.readthedocs.io/en/latest/); this might help.
> 
> Oh, thanks for the kind reference! Let's decide the specific solution to @kangz12345.

You mean that `ast` does not support 3.8, right? Then let's remove 3.8 from our requirements, i.e., support from 3.9.
However, when we have to deal with the compatibility issue in the future, `astunparse` would be very helpful 👍 

@BECATRUE Please make an issue for removing 3.8 from our GitHub actions 😄

main.py
```diff
@@ -146,15 +147,16 @@ async def request_termination_of_experiment(rid: int):
 
 
 def add_tracking_line(stmt_list: List[ast.stmt]) -> List[ast.stmt]:
-    """Modifies the given statements list to leave rtio logs that track each line.
+    """Returns the statements list created to leave rtio logs that track each line.
```
kangz12345: How about: `Returns a new statement list interleaved with rtio logs for tracking each line.`

kangz12345 (changes requested):

kangz12345:
> Are these errors of `result_path` in the unit tests scheduled for another issue/PR?

Maybe relevant test codes should be updated as well! @BECATRUE

main.py
BECATRUE: Okay! A user will submit the experiment through IQUIP and the GUI checks if the selected file is a valid experiment, i.e. if it contains the experiment class.
Thus, the experiment class always exists.

The problem is that `run()` doesn't exist, but this will not happen unless the user intends it, so I will only mark it in the docstring.

main.py
BECATRUE: Wow, I see. I will apply your idea!

main.py
kangz12345: Good! As you said, to guarantee such condition, all the other parts should work well and the mechanism should stay the same (what if we change the structure later?).
To decouple the components, we should think each part independently as possible.

BECATRUE:
> Are these errors of result_path in the unit tests scheduled for another issue/PR?

Yes! This PR is already so long that I plan to update the unit tests later.
@kangz12345 Is it okay to handle it in the later issue?

BECATRUE:
> You mean that ast does not support 3.8, right? Then let's remove 3.8 from our requirements, i.e., support from 3.9.

Exactly, `ast` supports 3.8, but `ast.unparse()` doesn't do that.
Using `astunparse` may be a good solution, but it looks complicated and I think remove 3.8 is enoughly good one!

giwon2004 (dismissed):
LGTM!

kangz12345:
I think the code has not been updated.

BECATRUE:
> I think the code has not been updated.

Now, I update it! Thanks for notification. @kangz12345

kangz12345 (approved):

giwon2004 (approved):

Issue #52, opened 2023-08-18
Release v1.2.0


BECATRUE:

kangz12345:
Let's release v1.2.0 after resolving #123 and implementing remote dataset viewer support.

BECATRUE:
I see. I will try to resolve #123 within this week.

kangz12345:
We released v2.0.0.

PR #53, opened 2023-08-18
Release v1.2.0

BECATRUE:
This is just a merge from `develop` branch to `release` one.

See also: #52


giwon2004:
I think #54 has to be also on this list, as it resolves an error made in #41.

BECATRUE:
> I think #54 has to be also on this list, as it resolves an error made in #41.

Oh, I see. Then, I will close this PR and let's reopen it when we're ready.

PR #54, opened 2023-08-19
Unit test for scheduler-related proxy request

### LINKED ISSUE #45: Unit test for fuctions connected to `SchedulerApp`
I will make a unit test for functions connected (from #44 and possibly #29) to `SchedulerApp`.

giwon2004:
It tests functions merged in #41 and #46.

An error was found while testing the functions, which is also resolved in this pull request.

This closes #45.

### DIFF
--- a/main.py
+++ b/main.py
@@ -6,6 +6,7 @@ import os
 import posixpath
 import shutil
 import time
+import copy
 import asyncio
 from contextlib import asynccontextmanager
 from datetime import datetime
@@ -112,7 +113,7 @@ async def get_experiment_queue() -> Dict[int, Dict[str, Any]]:
         It includes the running experiment with different "status" value.
     """
     remote = get_client("master_schedule")
-    current_queue = previous_queue
+    current_queue = copy.deepcopy(previous_queue)
     while current_queue == previous_queue:
         await asyncio.sleep(0)
         current_queue.clear()
--- a/test.py
+++ b/test.py
@@ -4,6 +4,7 @@ import json
 import logging
 import posixpath
 import time
+import copy
 import unittest
 from datetime import datetime
 from unittest import mock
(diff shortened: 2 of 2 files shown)


main.py
```diff
@@ -112,7 +113,7 @@ async def get_experiment_queue() -> Dict[int, Dict[str, Any]]:
         It includes the running experiment with different "status" value.
     """
     remote = get_client("master_schedule")
-    current_queue = previous_queue
+    current_queue = copy.deepcopy(previous_queue)
```
BECATRUE, line 116: 👍

test.py
```diff
@@ -60,6 +60,41 @@ def test_get_experiment_info(self):
...
+                    'arguments': {'user': 'QuIQCL', 'time': 1.0, 'save': False, 'color': 'r'},
+                    'file': 'DIRECTORY-PATH'
+                },
+                'priority': 1,
+                'due_date': None,
+                'flush': False,
+                'status': status,
+                'repo_msg': None
+            }
+        } for status in ['pending', 'preparing', 'running', 'run_done', 'analyzing', 'deleting'])
```
BECATRUE: I think the indentation looks not legible.
How about this?

```python
test_queues = tuple({
  '1': {
    'pipeline': 'main',
    'expid': {
      'log_level': 30,
      'class_name': None,
      'arguments': {'user': 'QuIQCL', 'time': 1.0, 'save': False, 'color': 'r'},
      'file': 'DIRECTORY-PATH'
    },
    'priority': 1,
    'due_date': None,
    'flush': False,
    'status': status,
    'repo_msg': None
  }} for status in ['pending', 'preparing', 'running', 'run_done', 'analyzing', 'deleting']
)
```

test.py
```diff
@@ -60,6 +60,41 @@ def test_get_experiment_info(self):
...
+        with TestClient(main.app) as client:
+            for queue in test_queues:
+                self.mocked_client.get_status.return_value = queue
+                response = client.get("/experiment/queue/")
+                self.assertEqual(response.status_code, 200)
+                self.assertEqual(response.json(), queue)
+
+    def test_delete_experiment(self):
+        test_rid = 2**31 - 1
```
BECATRUE: Is there any reason to set `test_rid` to a big number?

BECATRUE (changes requested):

test.py
giwon2004: Anything is okay. Any suggestions?

BECATRUE: There is no problem! But I think a small interger looks better e.g., 0 or 1. @giwon2004

test.py
kangz12345: If you cannot make it in a single line, do not use comprehension.
The longest acceptable form:
```python
generator = (
    long_expression
    for variable in iterable
    if condition
)
```
If the expression does not fit in a single line, please consider using a `for` statement, etc.

kangz12345: In addition, please use double quotes (`"`) for consistency.

kangz12345 (changes requested):

test.py
```diff
@@ -61,21 +61,24 @@ def test_get_experiment_info(self):
...
-                'priority': 1,
-                'due_date': None,
-                'flush': False,
-                'status': status,
-                'repo_msg': None
-            }
-        } for status in ['pending', 'preparing', 'running', 'run_done', 'analyzing', 'deleting'])
+        test_queues = tuple()
+        for status in ["pending", "preparing", "running", "run_done", "analyzing", "deleting"]:
+            test_queues += ({
```
kangz12345: Well, this is a performance or style issue, but `tuple` is designed for an immutable iterable.
If you want to mutate a data structure, `list` is preferred, i.e., `list.append()`.

In more detail, `tuple +` will make a new `tuple` object, so it will allocate a new memory space for each tuple addition (and the new space increases: 0, 1, 2, ...).
On the other hand, `list` will allocate some extra memory space in advance, so new memory allocations occur less frequently.

If you want to use `tuple`, you can wrap it at the end, but I don't think it is necessary.

Furthermore, I think you can just directly iterate over `status` instead of `queue` in the `with` block below. Then you don't have to define `test_queues`.

Please consider these idea and tell me your opinions!

kangz12345 (approved):
LGTM

BECATRUE (approved):
LGTM!

PR #58, opened 2023-09-01
Remove Python 3.8

### LINKED ISSUE #56: Remove Python 3.8
As we discussed in #50, `ast.unparse()` doesn't support Python 3.8 so we will remove it.

> > > How about supporting only after Python 3.8?
> > 
> > 
> > Check out [`astunparse`](https://astunparse.readthedocs.io/en/latest/); this might help.
> 
> Oh, thanks for the kind reference! Let's decide the specific solution to @kangz12345.

You mean that `ast` does not support 3.8, right? Then let's remove 3.8 from our requirements, i.e., support from 3.9.
However, when we have to deal with the compatibility issue in the future, `astunparse` would be very helpful 👍 

@BECATRUE Please make an issue for removing 3.8 from our GitHub actions 😄

_Originally posted by @kangz12345 in https://github.com/snu-quiqcl/artiq-proxy/issues/50#issuecomment-1690986853_

BECATRUE:
This PR will close #56.

I removed Python 3.8 from GitHub tests.

### DIFF
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.9", "3.10", "3.11"]
     steps:
     - uses: actions/checkout@v3
     - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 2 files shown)


BECATRUE:
The unittest will fail because I didn't update it.

kangz12345 (approved):
LGTM

giwon2004 (approved):

PR #59, opened 2023-09-02
Send the result directory structure

### LINKED ISSUE #38: Send the result directory structure

BECATRUE:
This PR will close #38.

# What I implemented
- Organize each RID directory
  - When the experiment is submitted, it makes a RID directory, `_{RID}/`
  - After organizing, rename the RID directory to `{RID}/` (remove the leading underscore)
- Send the RID list

# How to fetch the result directory structure
```shell
curl -X GET 127.0.0.1:8000/result/
```

# Result

### DIFF
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -21,6 +21,7 @@ jobs:
         pip install pylint
     - name: Add dependencies about external libraries
       run: |
+        pip install h5py
         pip install fastapi
         pip install git+https://github.com/m-labs/sipyco.git
     - name: Analyze the code with pylint
--- a/main.py
+++ b/main.py
@@ -315,7 +315,7 @@ async def get_running_experiment() -> Optional[int]:
     return None
 
 
-def organize_result_directory(result_dir_path: str, rid: str):
+def organize_result_directory(result_dir_path: str, rid: str) -> bool:
     """Organizes the result directory.
     
     It performs the following:
@@ -327,22 +327,20 @@ def organize_result_directory(result_dir_path: str, rid: str):
     Args:
         result_dir_path: The full path of the result directory.
         rid: The RID string.
+
+    Returns:
+        True if the h5 result file exists.
+        Otherwise, stop organizing the result directory and return False. 
     """
     rid_dir_path = posixpath.join(result_dir_path, f"{rid}/")
     # read and remove the metadata
     metadata_path = posixpath.join(rid_dir_path, "metadata.json")
     with open(metadata_path, encoding="utf-8") as metadata_file:
         metadata = json.load(metadata_file)
-    os.remove(metadata_path)
     submission_time_str, visualize = metadata["submission_time"], metadata["visualize"]
     submission_time = datetime.fromisoformat(submission_time_str)
     date = submission_time.date().isoformat()
     hour = submission_time.hour
-    # move the modified experiment file to the RID directory
-    if visualize:
-        experiment_path = posixpath.join(result_dir_path, f"experiment_{submission_time_str}.py")
-        moved_experiment_path = posixpath.join(rid_dir_path, "modified_experiment.py")
-        shutil.move(experiment_path, moved_experiment_path)
     # find the result file
     datetime_result_dir_path = posixpath.join(result_dir_path, f"{date}/{hour}/")
     padded_rid = rid.zfill(9)
@@ -350,7 +348,7 @@ def organize_result_directory(result_dir_path: str, rid: str):
         if item.startswith(padded_rid):
             break
     else:  # no result file
-        return
+        return False
     # copy the result file to the RID directory
     result_path = posixpath.join(datetime_result_dir_path, item)
     copied_result_path = posixpath.join(rid_dir_path, "result.h5")
@@ -363,6 +361,14 @@ def organize_result_directory(result_dir_path: str, rid: str):
         submission_time_dataset[0] = submission_time_str
         visualize_dataset = result_file.create_dataset("visualize", (1,), dtype="bool")
         visualize_dataset[0] = visualize
+    # move the modified experiment file to the RID directory
+    if visualize:
+        experiment_path = posixpath.join(result_dir_path, f"experiment_{submission_time_str}.py")
+        moved_experiment_path = posixpath.join(rid_dir_path, "modified_experiment.py")
+        shutil.move(experiment_path, moved_experiment_path)
+    # remove the metadata file
+    os.remove(metadata_path)
+    return True
 
 
 @app.get("/result/")
@@ -381,9 +387,11 @@ async def list_result_directory() -> List[int]:
         item_path = posixpath.join(result_dir_path, item)
         if posixpath.isdir(item_path) and item.isdigit():  # find a RID directory
             rid = int(item)
-            if rid > last_rid:
-                organize_result_directory(result_dir_path, item)
-            rid_list.append(rid)
+            if rid <= last_rid:  # alraedy organized
+                rid_list.append(rid)
+                continue
+            if organize_result_directory(result_dir_path, item):
+                rid_list.append(rid)
     # update the most recently fetched RID
     rid_list.sort()
     last_rid = rid_list[-1] if rid_list else -1
--- a/main.py
+++ b/main.py
@@ -337,8 +337,11 @@ def organize_result_directory(result_dir_path: str, rid: str) -> bool:
     rid_dir_path = posixpath.join(result_dir_path, f"{rid}/")
     # read the metadata
     metadata_path = posixpath.join(rid_dir_path, "metadata.json")
-    with open(metadata_path, encoding="utf-8") as metadata_file:
-        metadata = json.load(metadata_file)
+    try:
+        with open(metadata_path, encoding="utf-8") as metadata_file:
+            metadata = json.load(metadata_file)
+    except OSError:
+        return False
     submission_time_str, visualize = metadata["submission_time"], metadata["visualize"]
     submission_time = datetime.fromisoformat(submission_time_str)
     date = submission_time.date().isoformat()
--- a/main.py
+++ b/main.py
@@ -254,7 +254,9 @@ async def submit_experiment(  # pylint: disable=too-many-arguments, too-many-loc
     submission_time = datetime.now().isoformat(timespec="seconds")
     result_dir_path = posixpath.join(configs["master_path"], configs["result_path"])
     if visualize:
-        src_experiment_path = posixpath.join(configs["master_path"], configs["repository_path"], file)
+        src_experiment_path = posixpath.join(
+            configs["master_path"], configs["repository_path"], file
+        )
         modified_experiment_path = posixpath.join(
             result_dir_path,
             f"experiment_{submission_time}.py"
@@ -280,7 +282,7 @@ async def submit_experiment(  # pylint: disable=too-many-arguments, too-many-loc
     remote = get_client("master_schedule")
     rid = remote.submit(pipeline, expid, priority, due_date, False)
     # make the RID directory
-    rid_dir_path = posixpath.join(result_dir_path, f"{rid}/")
+    rid_dir_path = posixpath.join(result_dir_path, f"_{rid}/")
     os.makedirs(rid_dir_path)
     # save the metadata
     metadata = {
@@ -388,29 +390,18 @@ async def list_result_directory() -> List[int]:
         A list with RIDs of the submitted experiments, sorted in an ascending order.
     """
     rid_list = []
-    # read the most recently fetched RID
     result_dir_path = posixpath.join(configs["master_path"], configs["result_path"])
-    last_rid_path = posixpath.join(result_dir_path, "rid.json")
-    try:
-        with open(last_rid_path, encoding="utf-8") as last_rid_file:
-            last_rid = json.load(last_rid_file)
-    except FileNotFoundError:
-        last_rid = -1
     # navigate to each RID directory
     for item in os.listdir(result_dir_path):
         item_path = posixpath.join(result_dir_path, item)
-        if posixpath.isdir(item_path) and item.isdigit():  # find a RID directory
-            rid = int(item)
-            if rid <= last_rid:  # already organized
-                rid_list.append(rid)
-                continue
+        # find a RID directory not organized yet
+        if posixpath.isdir(item_path) and item.startswith("_") and item[1:].isdigit():
+            rid = item[1:]
+            rid_dir_path = posixpath.join(result_dir_path, f"{rid}/")
+            shutil.move(item_path, rid_dir_path)
             if organize_result_directory(result_dir_path, item):
                 rid_list.append(rid)
-    # update the most recently fetched RID
     rid_list.sort()
-    last_rid = rid_list[-1] if rid_list else -1
-    with open(last_rid_path, "w", encoding="utf-8") as last_rid_file:
-        json.dump(last_rid, last_rid_file)
     return rid_list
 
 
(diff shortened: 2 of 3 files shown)


main.py
```diff
@@ -1,6 +1,7 @@
 """Proxy server to communicate a client to ARTIQ."""
 
 import ast
+import h5py
```
kangz12345: I think this is not a standard library. Please move it to the external library imports.
Also, please check out the GitHub action errors. Maybe `h5py` should be added to the dependency.

main.py
```diff
@@ -313,6 +315,100 @@ async def get_running_experiment() -> Optional[int]:
...
 
 
+def organize_result_directory(result_dir_path: str, rid: str) -> bool:
+    """Organizes the result directory.
+    
+    It performs the following:
+        1. Read the metadata file from the RID directory and remove it.
+        2. Move the modified experiment to the RID directory.
+        3. Find the h5 result file and copy it to the RID directory.
+        4. Dump the metadata on the copied result file.
```
kangz12345: As you did in the actual code, removing the metadata file should go to the end.
Please reflect that to this docstring.

main.py
```diff
@@ -313,6 +315,100 @@ async def get_running_experiment() -> Optional[int]:
...
+    Args:
+        result_dir_path: The full path of the result directory.
+        rid: The RID string.
+
+    Returns:
+        True if the h5 result file exists.
+        Otherwise, stop organizing the result directory and return False. 
+    """
+    rid_dir_path = posixpath.join(result_dir_path, f"{rid}/")
+    # read and remove the metadata
```
kangz12345: Please get rid of "and remove" since it does not remove the metadata file here.

main.py
```diff
@@ -313,6 +315,100 @@ async def get_running_experiment() -> Optional[int]:
...
+    datetime_result_dir_path = posixpath.join(result_dir_path, f"{date}/{hour}/")
+    padded_rid = rid.zfill(9)
+    for item in os.listdir(datetime_result_dir_path):
+        if item.startswith(padded_rid):
+            break
+    else:  # no result file
+        return False
+    # copy the result file to the RID directory
+    result_path = posixpath.join(datetime_result_dir_path, item)
+    copied_result_path = posixpath.join(rid_dir_path, "result.h5")
```
kangz12345: The term "copied" is quite ambiguous: if you copy A and get B, we can say both A and B are _copied_.
Actually I think it is more common to say that "A is copied" and "B is a copy of A".

Therefore, I would suggest `new_result_path` for example. Or, you can use "src" and "dst" which are popular terms in computer science (`shutil` documentation also uses those terms).

main.py
```diff
@@ -313,6 +315,100 @@ async def get_running_experiment() -> Optional[int]:
...
+        submission_time_dataset = result_file.create_dataset(
+            "submission_time", (1,), dtype=h5py.string_dtype(encoding="utf-8")
+        )
+        submission_time_dataset[0] = submission_time_str
+        visualize_dataset = result_file.create_dataset("visualize", (1,), dtype="bool")
+        visualize_dataset[0] = visualize
+    # move the modified experiment file to the RID directory
+    if visualize:
+        experiment_path = posixpath.join(result_dir_path, f"experiment_{submission_time_str}.py")
+        moved_experiment_path = posixpath.join(rid_dir_path, "modified_experiment.py")
```
kangz12345: Here as well, "moved" can be interpreted as both meanings.

main.py
```diff
@@ -313,6 +315,100 @@ async def get_running_experiment() -> Optional[int]:
...
+
+    Returns:
+        True if the h5 result file exists.
+        Otherwise, stop organizing the result directory and return False. 
+    """
+    rid_dir_path = posixpath.join(result_dir_path, f"{rid}/")
+    # read and remove the metadata
+    metadata_path = posixpath.join(rid_dir_path, "metadata.json")
+    with open(metadata_path, encoding="utf-8") as metadata_file:
+        metadata = json.load(metadata_file)
```
kangz12345: I think we should handle the case where the metadata file does not exist.

main.py
```diff
@@ -313,6 +315,100 @@ async def get_running_experiment() -> Optional[int]:
...
+    # navigate to each RID directory
+    for item in os.listdir(result_dir_path):
+        item_path = posixpath.join(result_dir_path, item)
+        if posixpath.isdir(item_path) and item.isdigit():  # find a RID directory
+            rid = int(item)
+            if rid <= last_rid:  # alraedy organized
+                rid_list.append(rid)
+                continue
+            if organize_result_directory(result_dir_path, item):
+                rid_list.append(rid)
```
kangz12345: Is it common to have no result file? Don't we need to notify it, e.g., by logging?

kangz12345 (changes requested):

giwon2004:
Why does the unit test say `No module named 'h5py'`? Will you resolve this later?

main.py
```diff
@@ -313,6 +315,100 @@ async def get_running_experiment() -> Optional[int]:
...
+    
+    It performs the following:
+        1. Read the metadata file from the RID directory and remove it.
+        2. Move the modified experiment to the RID directory.
+        3. Find the h5 result file and copy it to the RID directory.
+        4. Dump the metadata on the copied result file.
+
+    Args:
+        result_dir_path: The full path of the result directory.
+        rid: The RID string.
```
giwon2004: How about `rid: The run identifier value of the experiment in string.`?

main.py
```diff
@@ -313,6 +315,100 @@ async def get_running_experiment() -> Optional[int]:
...
+        with open(last_rid_path, encoding="utf-8") as last_rid_file:
+            last_rid = json.load(last_rid_file)
+    except FileNotFoundError:
+        last_rid = -1
+    # navigate to each RID directory
+    for item in os.listdir(result_dir_path):
+        item_path = posixpath.join(result_dir_path, item)
+        if posixpath.isdir(item_path) and item.isdigit():  # find a RID directory
+            rid = int(item)
+            if rid <= last_rid:  # alraedy organized
```
giwon2004: Small typo: `already` 😅

main.py
```diff
@@ -313,6 +315,100 @@ async def get_running_experiment() -> Optional[int]:
...
+                rid_list.append(rid)
+                continue
+            if organize_result_directory(result_dir_path, item):
+                rid_list.append(rid)
+    # update the most recently fetched RID
+    rid_list.sort()
+    last_rid = rid_list[-1] if rid_list else -1
```
giwon2004: If the experiments are:
```
rid = 1, priority = 3
rid = 2, priority = 5
```

Then the `rid = 2` experiment will be run first, which makes the `rid = 1` experiment the most recent experiment that ran.
Did you intend `last_rid = 2`? If so, why is sorting (in the line above) needed?

giwon2004 (changes requested):

main.py
BECATRUE: Oh, thanks for notifying me!

main.py
BECATRUE: I see. I will handle the exception.

main.py
BECATRUE: It may occur commonly!

For example, you submit a long-time experiment and fetch the RID list right after.
When the experiment is submitted, the RID directory is created, but it will take a long time to create the result file.
In this case, it must not organize this RID directory until it finishes.

In summary, a running or queued experiment must not be organized!

main.py
BECATRUE: It makes more legible!

main.py
BECATRUE: Oh, I made a mistake.
I intended `last_rid = 1`, but I didn't consider that the actual order of execution of the experiment may differ from that of the RID.

BECATRUE:
I realized that there are several wrong implementations, so I will close this PR and repoen it when they are resolved.

main.py
```diff
@@ -313,6 +317,117 @@ async def get_running_experiment() -> Optional[int]:
...
+    src_result_path = posixpath.join(datetime_result_dir_path, item)
+    dst_result_path = posixpath.join(rid_dir_path, "result.h5")
+    shutil.copyfile(src_result_path, dst_result_path)
+    # modify the result file
+    with h5py.File(dst_result_path, "a") as result_file:
+        # start time
+        start_time = result_file["start_time"][()]
+        del result_file["start_time"]
+        start_time_str = datetime.fromtimestamp(start_time).isoformat()
+        start_time_dataset = result_file.create_dataset(
```
giwon2004, line 372: What is the difference between the original  `result_file["start_time"]` and this one?

main.py
```diff
@@ -313,6 +317,117 @@ async def get_running_experiment() -> Optional[int]:
...
+            if organize_result_directory(result_dir_path, rid):
+                rid_dir_path = posixpath.join(result_dir_path, f"{rid}/")
+                os.rename(item_path, rid_dir_path)
+    # find all organized RID directories
+    rid_list = []
+    for item in os.listdir(result_dir_path):
+        item_path = posixpath.join(result_dir_path, item)
+        if posixpath.isdir(item_path) and item.isdigit():
+            rid_list.append(int(item))
+    rid_list.sort()
```
giwon2004, line 430: So this sort means:
"Return all experiments that have been completed, in ascending order of rid."
Am I right?

giwon2004 (commented):
Simple questions!

main.py
BECATRUE: The original one is the same format as `time.time()` so it is just a float representing the time in seconds from the specific time stamp.
But, it looks not legible, so I replaced it with the iso format, e.g., 2023-09-05T12:00:00.

main.py
BECATRUE: Right! Is there any problem?

main.py
```diff
@@ -313,6 +317,117 @@ async def get_running_experiment() -> Optional[int]:
     return None
 
 
+# pylint: disable=too-many-locals
+def organize_result_directory(result_dir_path: str, rid: str) -> bool:
+    """Organizes the result directory.
+    
+    It performs the following:
+        1. Read the metadata file from the RID directory and remove it.
```
kangz12345: Please remove "and remove it."

main.py
```diff
@@ -313,6 +317,117 @@ async def get_running_experiment() -> Optional[int]:
...
+        Otherwise, stop organizing the result directory and return False. 
+    """
+    rid_dir_path = posixpath.join(result_dir_path, f"_{rid}/")
+    # read the metadata
+    metadata_path = posixpath.join(rid_dir_path, "metadata.json")
+    try:
+        with open(metadata_path, encoding="utf-8") as metadata_file:
+            metadata = json.load(metadata_file)
+    except OSError:  # no chance of happening
+        return False
```
kangz12345: If you handle such case like this, we cannot discriminate the two `False` cases: no metadata or no h5.
Moreover, we cannot expect users to act as we expected. Users have access to file system, so they can create, modify, and delete files and directories. Therefore, the program should be prepared for such unexpected environments, i.e., _exceptions_.

kangz12345: In my opinion, you should come up with a way to handle such two cases differently or at least log the exception.

main.py
giwon2004: Not at all.

giwon2004 (dismissed):
LGTM!

main.py
```diff
@@ -347,65 +354,77 @@ def organize_result_directory(result_dir_path: str, rid: str) -> bool:
...
-        if posixpath.isdir(item_path) and item.isdigit():  # find a RID directory
-            rid = int(item)
-            if rid <= last_rid:  # alraedy organized
-                rid_list.append(rid)
-                continue
-            if organize_result_directory(result_dir_path, item):
-                rid_list.append(rid)
-    # update the most recently fetched RID
+        if posixpath.isdir(item_path) and item.isdigit():
+            rid_list.append(int(item))
```
kangz12345: Isn't it better to do it within a single iteration?
I think this code is kind of duplicate.

kangz12345 (changes requested):

.github/workflows/pylint.yml
```diff
@@ -21,6 +21,7 @@ jobs:
         pip install pylint
     - name: Add dependencies about external libraries
       run: |
+        pip install h5py
```
kangz12345, line 24: Please add this to the unit test action as well.

kangz12345 (changes requested):

main.py
BECATRUE: I think leaving a log would be better because it looks inefficient to raise an exception without returning the RID list because of only one invalid file.

BECATRUE: So I came up with two ways.
1. Use `print()`: It is the simplist way, and it leave the logs in the original log file containing all request logs, too.
2. Use `logging`: It is similar to QIWIS. It creates another log file.

Which looks better? @kangz12345

.github/workflows/pylint.yml
BECATRUE: I planned this later becuase the unit test wasn't updated in this PR.
But you metioned it, so I will apply it.

main.py
BECATRUE: The returned values of `os.listdir(result_dir_path)` of each for loop may be different since a directory can be renamed in the first loop.
Thus, to merge them in one loop, there should be two cases:
1. The directory wasn't organized, but it is organized now.
2. The directory is already organized.

It may look more complicated, nevetheless is it okay?

kangz12345: The only change is removing the heading "_", right?
Is the following code invalid?
```python
rid_list = []
for item in os.listdir(result_dir_path):
    item_path = posixpath.join(result_dir_path, item)
    if not posixpath.isdir(item_path):
        continue
    if item.startswith("_") and item[1:].isdigit():
        rid = item[1:]
        if organize_result_directory(result_dir_path, rid):
            rid_dir_path = posixpath.join(result_dir_path, f"{rid}/")
            os.rename(item_path, rid_dir_path)
            rid_list.append(int(rid))
    elif item.isdigit():
        rid_list.append(int(item))
rid_list.sort()
```

.github/workflows/pylint.yml
kangz12345: It is not a problem that you postpone testing, but you should let reviewers know!

kangz12345: + Even if you planned it as TODO, I think it is necessary to make the unit test not failing.

main.py
kangz12345: Since this is an asynchronous server, I think `logging` would be more stable.
However, for now, you don't have to configure the logger setup since it is out of the scope of this PR.
You can just define `logger = logging.getLogger(__name__)` and use it.

main.py
kangz12345: As we discussed offline, please create an issue for using a subdirectory instead of the prefix "_" and ignore this review comment in this PR.

.github/workflows/pylint.yml
BECATRUE: Okay! Then, I will add it in this PR.

BECATRUE:
The unit tests will be resolved a few days later.

.github/workflows/pylint.yml
kangz12345: Hmm, the unit test is still failing, so I think you should inspect it here.
If it is the problem of the test (the test was already wrong), you can resolve it in another PR later.
If the test _was_ correct but it now fails because of this PR, I think you should resolve it here because it could "blind" us from other test failures (we might not see the other test failures since we know and expect that the test should fail).

kangz12345: As we discussed offline, the error has been occurred even before this PR, so it will be resolved in another PR.

BECATRUE:
Can you check this PR? @kangz12345

kangz12345 (approved):
LGTM

PR #65, opened 2023-09-18
Skip the unittest about submission temporarily

BECATRUE:
This closes #63.

I just added `@unittest.skip()` decorator to skip `test_submit_experiment()`.

PR #67, opened 2023-09-18
Set the override and level option of a specific TTL

### LINKED ISSUE #66: Set the override and level option of a specific TTL
_(updated)_

To control ARTIQ hardwares manually not through an experiment, we should be able to set the value of a specific TTL explicitly.
For this, in ARTIQ the function in terms of override and level is implemented in [moninj.py](https://github.com/m-labs/artiq/blob/master/artiq/dashboard/moninj.py#L701).

Similar to this, I will implement setting the override and level option of a specific TTL.

#### config.json
To limit the TTLs that users can control, a list, `ttl_channels` will be added to `config.json`.
Only TTL channel in `ttl_channels` can be overridden or controlled manually.

#### main.py
When receiving a request with the url, `~/ttl/level/?channel={channel}&value={value}`, it will set the level of the TTL channel `channel` to `value`.
When receiving a request with the url, `~/ttl/override/?value={value}`, it will set the override of all TTLs described in `config.json` to `value`.

BECATRUE:
This closes #66.

For details, please read #66.

To test in miniPC, you can use the following commands.
```shell
$ curl -X POST 'http://127.0.0.1:8000/ttl/level/?channel=23&value=True'  // set the level of channel 23 to HIGH
$ curl -X POST 'http://127.0.0.1:8000/ttl/override/?value=True'  // set the override of all channels to TRUE
```

### DIFF
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -24,6 +24,7 @@ jobs:
         pip install h5py
         pip install fastapi
         pip install git+https://github.com/m-labs/sipyco.git
+        pip install git+https://github.com/m-labs/artiq.git
     - name: Analyze the code with pylint
       run: |
         pylint main.py --report=y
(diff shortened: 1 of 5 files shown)


BECATRUE:
The unittest will pass after merging #65.

kangz12345 (approved):
LGTM

PR #71, opened 2023-09-21
Set the voltage of a specific DAC

### LINKED ISSUE #69: Set the voltage of a specific DAC
As the next step of #66, we need to set the voltage of a specific DAC.
There is no neat way to set not through an experiment, we will just submit an experiment to set the voltage.

#### config.json
To limit the DACs that users can control, a list `dac_devices` will be added to `config.json`.
Only DAC device in `dac_devices` can be set manually.

#### main.py
When receiving a request with the url, `~/dac/voltage/?device={device}&channel={channel}&value={value}`, it will set the voltage of the DAC to `value`.

BECATRUE:
This closes #69.

I didn't check through an oscilloscope, but checked that the submission works well and there is no error in artiq-master.

### How to send a query
```shell
# set the voltage of channel 1 in DAC zotino0 to 5V
curl -X POST 'http://127.0.0.1:8000/dac/voltage/?device="zotino0"&channel=1&value=5'
```

### DIFF
--- a/main.py
+++ b/main.py
@@ -524,14 +524,13 @@ from artiq.experiment import *
 class {class_name}(EnvExperiment):
     def build(self):
         self.setattr_device("core")
-        self.setattr_device("ttl23")
+        self.dac = self.get_device({device})
 
     @kernel
     def run(self):
         self.core.reset()
-        self.ttl23.on()
-        delay(2*s)
-        self.ttl23.off()
+        self.dac.init()
+        self.dac.set_dac([{value}], [{channel}])
 """
     expid = {
         "log_level": logging.WARNING,
--- a/main.py
+++ b/main.py
@@ -530,6 +530,7 @@ class {class_name}(EnvExperiment):
     def run(self):
         self.core.reset()
         self.dac.init()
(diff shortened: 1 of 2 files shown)


main.py
```diff
@@ -507,6 +508,42 @@ async def set_ttl_override(value: bool):
...
+class {class_name}(EnvExperiment):
+    def build(self):
+        self.setattr_device("core")
+        self.dac = self.get_device({device})
+
+    @kernel
+    def run(self):
+        self.core.reset()
+        self.dac.init()
+        delay(200*us)
```
kangz12345, line 533: What is this for? Is it just copied from artiq codes?

kangz12345 (approved):

main.py
BECATRUE: Unless inserting a delay, it cause an underflow exception.
So I found out the test code in ARTIQ GitHub and ARTIQ also inserts a delay in [artiq_sinara_tester.py](https://github.com/m-labs/artiq/blob/master/artiq/frontend/artiq_sinara_tester.py#L494).

PR #72, opened 2023-09-22
Read master dataset

### LINKED ISSUE #70: Read the broadcast datasets
Fetch the datasets that are broadcast to artiq master.

Artiq master will hold the datasets, so we will use RPC to call the getter method of the dataset DB.

### Additional context
Still investigating...

kangz12345:
This closes #70.

`get_master_dataset(key)` returns the dataset in the artiq master's dataset DB, searching the value for the given key.

### DIFF
--- a/main.py
+++ b/main.py
@@ -11,10 +11,11 @@ import copy
 import asyncio
 from contextlib import asynccontextmanager
 from datetime import datetime
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Optional, Union
 
 import h5py
 import pydantic
+import numpy as np
 from fastapi import FastAPI
 from sipyco import pc_rpc as rpc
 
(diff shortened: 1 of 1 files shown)


main.py
```diff
@@ -448,6 +449,20 @@ async def list_result_directory() -> List[int]:
...
+@app.get("/dataset/master/")
+async def get_master_dataset(key: str) -> Union[int, float, List]:
+    """Returns the dataset broadcast to artiq master.
+    
+    Args:
+        key: The key of the target dataset.
+    """
+    remote = get_client("master_dataset_db")
+    array = remote.get(key)
+    if isinstance(array, np.ndarray):
```
BECATRUE, line 461: It is just a question!
When isn't the `array` a numpy array? The case it is a number?

BECATRUE (approved):
I'm sorry that I checked too late. It looks cool!

main.py
kangz12345: The dataset in artiq does not restrict the type of the values in it. It can be anything in principle!

BECATRUE: Oh, I see! Thanks👍

PR #76, opened 2023-09-25
Use TTL name to set the override and level

### LINKED ISSUE #75: Use TTL name to set the override and level
As @kangz12345 and I discussed in #68, we decided to use TTL name, not TTL channel number similar to DAC.

### load_config_file()
When starting the server, it reads device DB file i.e., `device_db.py` and store the contents.

### set_ttl_level() and set_ttl_override()
Use TTL name as an argument, find its channel from device DB.

BECATRUE:
This closes #75.

To send a query, use the following commands:
```shell
# Set the level
$ curl -X POST 'http://127.0.0.1:8000/ttl/level/?device=ttl23&value=True'

# Set the override
$ curl -X POST 'http://127.0.0.1:8000/ttl/override/?value=True'
```

I tested if setting the level and override worked well.

### DIFF
--- a/main.py
+++ b/main.py
@@ -38,7 +38,7 @@ def load_config_file():
         "master_path": {master_path},
         "repository_path": {repository_path},
         "result_path": {result_path},
-        "ttl_channels": [{channel0}, {channel1}, ... ],
+        "ttl_devices": [{ttl_device0}, {ttl_device1}, ... ],
         "dac_devices": [{dac_device0}, {dac_device1}, ... ]
       }
     """
--- a/main.py
+++ b/main.py
@@ -1,6 +1,7 @@
 """Proxy server to communicate a client to ARTIQ."""
 
 import ast
+import importlib.util
 import json
 import logging
 import os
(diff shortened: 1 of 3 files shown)


main.py
```diff
@@ -38,14 +41,25 @@ def load_config_file():
...
 
 
+def load_device_db():
+    """Loads device DB from the device DB file."""
+    device_db_full_path = posixpath.join(configs["master_path"], configs["device_db_path"])
+    module_name = "device_db"
+    spec = importlib.util.spec_from_file_location(module_name, device_db_full_path)
+    module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(module)
+    device_db.update(module.device_db)
```
kangz12345, line 60: This seems okay now, but I think we should update this to use RPC to fetch the artiq master's device DB.

kangz12345 (approved):

main.py
BECATRUE: Okay! I created an issue in terms of this.

PR #80, opened 2023-09-26
Set the default profile of a specific DDS

### LINKED ISSUE #79: Set the default profile of a specific DDS
As the next step of #69, we need to set the default profile of a specific DDS.
For this, we will submit an experiment to do this.

#### config.json
To limit the DDSs that users can control, a list `dds_devices` will be added to `config.json`.
Only DDS device in `dds_devices` can be set manually.

#### main.py
When receiving a request with the url, `~/dds/profile/?device={device}&channel={channel}&freq={frequency}&amp={amplitude}&phase={phase}&switching={switching}`, it will set the default profile of the DDS to `frequency`, `amplitude`, and `phase`.
If `switching` is True, it will change the current profile into the default profile in order to apply the setting actually.

BECATRUE:
This closes #79.

~~I didn't test if it worked well, but will do it until today.~~
I tested it, and it works well!

For a request, refer to the following command.
```shell
$ curl -X POST 'http://127.0.0.1:8000/dds/profile/?device=urukul0&channel=1&frequency=200&amplitude=0.8&phase=0.4&switching=False'
```


main.py
```diff
@@ -549,6 +553,9 @@ async def set_dac_voltage(device: str, channel: int, value: float):
         channel: The DAC channel number. For Zotino, there are 32 channels, from 0 to 31.
         value: The voltage to set. For Zotino, the valid range is from -10V to +10V.
     """
+    if device not in configs["dac_devices"] or channel not in configs["dac_devices"][device]:
+        logger.exception("The DAC device %s CH %d is not defined in config.json.", device, channel)
```
kangz12345: Since this is not in the try-except statement, please use `error()` instead of `exception()`.

main.py
```diff
@@ -576,6 +583,57 @@ def run(self):
...
+    Args:
+        device: The DDS device name described in device_db.py.
+        channel: The DDS channel number. For Urukul, there are 4 channels, from 0 to 3.
+        frequency: The frequency to set. For Urukul, the valid range is from 1HHz to 400MHz.
+        amplitude: The amplitude to set. For Urukul, the valid range is from 0 to 1.
+        phase: The phase to set. For Urukul, the valid range is from 0 to 1.
+        switching: If True, the current profile is switched to the default profile.
+    """
+    if device not in configs["dds_devices"] or channel not in configs["dds_devices"][device]:
+        logger.exception("The DDS device %s CH %d is not defined in config.json.", device, channel)
```
kangz12345: `error()` instead of `exception()`!

kangz12345 (changes requested):

kangz12345 (approved):

PR #86, opened 2023-10-09
Fix that fetching the scheduled queue works wrong

### LINKED ISSUE #85: Fetching the scheduled queue works wrong when there are several clients
### Describe the bug
When several clients (in here, I assumes there are only two clients) try to fetch the scheduled queue, each event such as one experiment becomes "preparing" status, is removed, etc. will be detected by ONLY ONE client.

Let's follow an example.

In two terminals, they are waiting the response of `curl -X GET 'http://127.0.0.1:8000/experiment/queue/'` request.
As soon as an experiment is submitted, one of terminals detects its change and updates `previous_queue` to the latest queue.
Thus, the other terminal CANNOT detect it, so it doesn't fetch the response.

image

In the above screenshot, one detected the change of status "preparing", but the other deteced the next event, that of status "running".

### Expected behavior
All clients should fetch all events.

### How to fix
As I guess, in while loop, using the copy of `previous_queue` instead of `previous_queue` directly may resolve this bug.

BECATRUE:
This closes #85.
For details, please see the issue contents.

As I tested with two proxy clients, it seems to work well for all cases.
Also, I made the unittest skip `get_experiment_queue()` and I will handle it later.

### How to test
```shell
# it returns immediately
$ curl -X GET 'http://127.0.0.1:8000/experiment/queue/'
{"updated_time":1696841009.986156,"queue":{}}

# it polls until the schedule is updated, and then return it
$ curl -X GET 'http://127.0.0.1:8000/experiment/queue/?updated_time=1696841009.986156'
{"updated_time":1696841093.070275,"queue":{"2709":{"pipeline":"main","expid":{"log_level":30,"class_name":null,"arguments":{"time":0.01,"count":1000},"file":"/home/rabi/artiq/repository/test/TTL/TTL_wave.py"},"priority":0,"due_date":null,"flush":false,"status":"preparing","repo_msg":null}}}
```

### DIFF
--- a/main.py
+++ b/main.py
@@ -90,6 +90,13 @@ def load_device_db():
     device_db.update(module.device_db)
 
 
+def init_schedule():
+    """Initializes the schedule info to the latest."""
+    remote = get_client("master_schedule")
+    queue = remote.get_status()
+    latest_schedule.update(queue)
+
+
 async def connect_moninj():
     """Creates a CommMonInj instance and connects it to ARTIQ."""
     def do_nothing(*_):
@@ -107,6 +114,7 @@ async def lifespan(_app: FastAPI):
     """
     load_configs()
     load_device_db()
+    init_schedule()
     await connect_moninj()
     yield
     await mi_connection.close()
--- a/main.py
+++ b/main.py
@@ -1,7 +1,6 @@
 """Proxy server to communicate a client to ARTIQ."""
 
 import ast
-import dataclasses
 import importlib.util
 import json
 import logging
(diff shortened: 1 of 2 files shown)


main.py
```diff
@@ -142,28 +174,29 @@ async def get_experiment_info(file: str) -> Any:
...
+        it returns the latest schedule info immediately.
+        Otherwise, it polls until the schedule is updated and then returns it.
     """
+    if updated_time is None or updated_time < latest_schedule.updated_time:
+        return latest_schedule
     remote = get_client("master_schedule")
-    current_queue = copy.deepcopy(previous_queue)
-    while current_queue == previous_queue:
+    latest_queue = copy.deepcopy(latest_schedule.queue)
+    queue = copy.deepcopy(latest_queue)
```
kangz12345: How about `queue = remote.get_status()`?

BECATRUE: Wow, it looks much better👍

kangz12345 (approved):
LGTM

Issue #87, opened 2023-10-13
Use `Subscriber` to _watch_ dataset modifications and provide dataset fetching API


kangz12345:
### Feature you want to implement
The proxy server will watch all the broadcast dataset modifications and keep it in its own data structure.
New API will be introduced:
- Get currently available dataset names
- Get modifications on a specific dataset since specific time stamp

### How the feature is implemented
As I tested, `Subscriber`'s `notify_cb()` is called with an argument `mod`.
`mod` is a `dict` with keys "action", "path", "x", "i", etc.
For example, when we call `append_to_dataset("name", value)`, then `mod={"action": "append", "path": ["name", 1], "x": value}`. (I'm not sure what the `1` in `"path"` is.)

Using this information, the `artiq-proxy` server will manage a dictionary (key=dataset_name, value=modification_list).
In addition, each modification is saved with its modified time stamp.

Furthermore, since the datasets can be reset, we should save the last-reset time stamps as well.
Imagine the case where a user requested modifications before and after the dataset is reset.

BECATRUE:
You may already know, but I think it'll help, so leave it.
In [sipyco.sync_struct](https://github.com/m-labs/sipyco/blob/master/sipyco/sync_struct.py), the modification dictionary spec is defined.

kangz12345:
> You may already know, but I think it'll help, so leave it.
> In [sipyco.sync_struct](https://github.com/m-labs/sipyco/blob/master/sipyco/sync_struct.py), the modification dictionary spec is defined.

That helped me a lot! I did not read it thoroughly before. Thanks 👍

PR #88, opened 2023-10-16
`DatasetTracker`

kangz12345:
This implements helper classes for #87.

There are some design issues:
1. `Modification` is currently just `Dict[str, Any]`. Should we impose more data structure guide on this? E.g., `dataclass`, `namedtuple`, etc.
2. `DatasetTracker.since()` creates an entry if the given dataset name does not exist. Is this good?

Test example:
```python
from dataset import *

tracker = DatasetTracker(maxlen=3)
tracker.add("test", 0.5, {"action": "append", "x": 0})
tracker.add("test", 1.0, {"action": "append", "x": 1})
tracker.add("test", 1.5, {"action": "append", "x": 2})
last, mods = tracker.since("test", 0.7)
print(last)  # 1.5
print(mods)  # ({'action': 'append', 'x': 1}, {'action': 'append', 'x': 2})
tracker.add("test", 2.0, {"action": "append", "x": 3})  # remove the first item (over max)
last, mods = tracker.since("test", 0.0)
print(last)  # 2.0
print(mods)  # ({'action': 'append', 'x': 1}, {'action': 'append', 'x': 2}, {'action': 'append', 'x': 3})
last, mods = tracker.since("test", 2.0)  # exclusive
print(last)  # 2.0
print(mods)  # ()
last, mods = tracker.since("empty", 1.0)  # create "empty" by defaultdict
print(last)  # 1.0
print(mods)  # ()
print(tracker.datasets())  # ('test', 'empty')
```

### DIFF
--- a/dataset.py
+++ b/dataset.py
@@ -61,3 +61,7 @@ class DatasetTracker:
 
     def __init__(self):
         self._modifications: Dict[str, ModificationQueue] = {}
+
+    def datasets(self) -> Tuple[str]:
+        """Returns the available dataset names."""
+        return tuple(self._modifications)
--- a/dataset.py
+++ b/dataset.py
@@ -53,8 +53,11 @@ class SortedQueue(Generic[K, V]):
         return latest_key, values
 
 
+ModificationQueue = SortedQueue[float, Dict[str, Any]]
+
+
 class DatasetTracker:
     """Holds dataset modifications and provides searching API."""
 
     def __init__(self):
-        self._modifications: Dict[str, SortedQueue[float, Dict[str, Any]]]
+        self._modifications: Dict[str, ModificationQueue] = {}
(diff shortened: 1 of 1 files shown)


BECATRUE:
> Modification is currently just Dict[str, Any]. Should we impose more data structure guide on this? E.g., dataclass, namedtuple, etc.

I think it's necessary one day, but it's okay now.

BECATRUE:
> DatasetTracker.since() creates an entry if the given dataset name does not exist. Is this good?

I don't like it, but as I understand, we won't use `DatasetTracker` directly, so it looks okay.

BECATRUE (approved):
LGTM!

BTW, did you intend that using this, the server sends only modification of datasets, not all?

kangz12345:
Then I will change implementation later when I think it is necessary.

kangz12345:
> BTW, did you intend that using this, the server sends only modification of datasets, not all?

Yes, because there is already a method to get the whole dataset. I will modify it to return the "updatedTime" as well.

PR #90, opened 2023-10-17
Subscribe dataset modification

kangz12345:
This PR includes:
1. New config file entries for the dataset tracker.
2. `dataset` module update (API change)
3. Use `Subscriber` to receive the target changes asynchronously along with the `fastapi` app.
4. Add an app API for getting modifications since a certain timestamp (tested).
5. Mock `dataset_tracker` and `init_dataset_tracker()` in `test` module to avoid unit test failure. Did not implement tests for the new features (should do this later)!

Does not include:
- Last deletion timestamp tracking
- Long-polling API (the current implementation returns immediately)

Known limits:
- Since `"path"` information is popped, this does not support more complex modifications with `"path"` information, e.g., nested dict, etc. I think we can implement it later if such a feature is necessary.

See also: #87


dataset.py
```diff
@@ -85,16 +108,47 @@ def add(self, dataset: str, timestamp: float, modification: Modification):
...
 
-    def since(self, dataset: str, timestamp: float) -> Tuple[float, Tuple[Modification]]:
+    def since(self, dataset: str, timestamp: float) -> Tuple[float, Tuple[Modification, ...]]:
         """Returns the latest timestamp and modifications since the given timestamp.
 
         See SortedQueue.tail() for details.
         
         Args:
             dataset: Target dataset name.
+              If dataset does not exist, it returns (-1, ()).
```
BECATRUE: How about moving this description and that of `timestamp` into "Returns:" section?

main.py
```diff
@@ -1,19 +1,20 @@
...
 import posixpath
 import shutil
 import time
-import copy
-import asyncio
 from contextlib import asynccontextmanager
 from datetime import datetime
 from enum import Enum
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Dict, List, Tuple, Optional, Union
```
BECATRUE: It is too minor, but could you exchange "Tuple" and "Optional" to sort correctly?

main.py
```diff
@@ -65,12 +71,18 @@ def load_configs():
...
         "result_path": {result_path},
+        "core_addr": {core_ip},
+        "master_addr": {artiq_master_ip},
         "device_db_path": {device_db_path},
         "ttl_devices": [{ttl_device0}, {ttl_device1}, ... ],
         "dac_devices": {
             {dac_device0}: [{dac_device0_channel0}, {dac_device0_channel1}, ... ],
             {dac_device1}: [{dac_device1_channel0}, {dac_device1_channel1}, ... ],
             ...
+        },
```
BECATRUE, line 82: It was my mistake, but could you write "dds_devices" structure?

main.py
```diff
@@ -516,6 +555,25 @@ async def get_master_dataset(key: str) -> Union[int, float, List]:
...
+
+@app.get("/dataset/master/modification/")
+async def get_dataset_modification(
+    key: str, timestamp: float
+) -> Tuple[float, Tuple[dset.Modification, ...]]:
+    """Returns the dataset modifications since the given timestamp.
+    
+    Args:
+        key: The key of the target dataset.
+        timestamp: The timestamp of the last update. See dataset.DatasetTracker.since().
```
BECATRUE: It looks better to move "See dataset..." above "Args:" since it is an overall explanation of this function.

main.py
```diff
@@ -95,11 +107,37 @@ def init_schedule():
...
+    global dataset_tracker  # pylint: disable=global-statement
+    maxlen = configs["dataset_tracker"].get("maxlen", 1 << 16)
+    dataset_tracker = dset.DatasetTracker(maxlen)
+    notify_cb = functools.partial(dset.notify_callback, dataset_tracker)
+    subscriber = Subscriber("datasets", lambda x: x, notify_cb)
+    await subscriber.connect(configs["master_addr"], configs["dataset_tracker"]["port"])
+    return asyncio.create_task(run_subscriber(subscriber))
+
+
+async def run_subscriber(subscriber: Subscriber):
```
BECATRUE: How about locating this above `init_dataset_tracker()`?

main.py
```diff
@@ -95,11 +107,37 @@ def init_schedule():
     latest_schedule.update(queue)
 
 
+async def init_dataset_tracker() -> asyncio.Task:
+    """Initializes the dataset tracker and runs the subscriber.
+    
+    This should be called after loading config.
+    """
+    global dataset_tracker  # pylint: disable=global-statement
+    maxlen = configs["dataset_tracker"].get("maxlen", 1 << 16)
```
BECATRUE, line 133: Is there any reason to set the default maxlen to "1 << 16"?

dataset.py
```diff
@@ -85,16 +108,47 @@ def add(self, dataset: str, timestamp: float, modification: Modification):
...
+    Args:
+        tracker: Target dataset tracket object.
+        mod: The argument of notify_cb() called by sipyco.sync_struct.Subscriber.
+    """
+    action = mod["action"]
+    if action == "init":
+        return
+    if not mod["path"]:
+        if action == "setitem":
+            tracker.add_dataset(mod["key"])
```
BECATRUE, line 148: Don't we need to set dataset contents?
According to [sync_struct.py](https://github.com/m-labs/sipyco/blob/c4f18adb658e3a546319425750cebeb8c88a016a/sipyco/sync_struct.py#L79), it adds a key, and also sets the value in the corresponding key.

BECATRUE (changes requested):

dataset.py
kangz12345: Since `DatasetTracker` only tracks modifications, I think it does not have to set the contents.
It only contains the modifications, and the user should call `get_master_dataset()` to get the dataset itself. Then the user will fetch modifications to update their dataset.
Do you understand what I mean?

main.py
kangz12345: I just wanted to set it to a power of 2, and I thought it should be greater than ~50,000 to store enough modifications.
I'm not sure yet how the size affects the performance.

dataset.py
kangz12345: I think the `timestamp` description should stay the same. It describes the role of the `timestamp` argument.

dataset.py
BECATRUE: I see! Then, there is no problem :)

BECATRUE (approved):
LGTM!

PR #92, opened 2023-10-18
Track last deletion timestamps

kangz12345:
By recording the last deletion timestamps, it can notify the client that the dataset has been deleted or reset.

See also: #87


BECATRUE (approved):
LGTM!

PR #94, opened 2023-10-18
Long polling for `get_dataset_modification()`

kangz12345:
This PR establishes a long polling system for `get_dataset_modification()`.

I tested some scenarios and it works well.
1. Immediately return `(-1, ())` when the dataset does not exist
2. Immediately return `(-1, ())` when the dataset has been deleted after the given timestamp
3. Immediately return valid values when there is modifications since the given timestamp
4. Wait for given timeout and return `(timestamp, ())` when the dataset exists and there was no modification since the given timestamp for given timeout
5. In case 4, if some modifications are added before the timeout, it returns valid values
6. In case 4, if the dataset is deleted or reset, it returns `(-1, ())`

See also: #87

### DIFF
--- a/main.py
+++ b/main.py
@@ -587,8 +587,7 @@ async def get_dataset_modification(
     if modifications or latest < 0 or timeout <= 0:
         return latest, modifications
     try:
-        async with asyncio.timeout(timeout):
-            await dataset_tracker.modified[key].wait()
+        await asyncio.wait_for(dataset_tracker.modified[key].wait(), timeout)
     except asyncio.TimeoutError:
         return latest, modifications
     return dataset_tracker.since(key, timestamp)
--- a/main.py
+++ b/main.py
@@ -584,7 +584,7 @@ async def get_dataset_modification(
           it can guarantee a valid return value for even short timeouts.
     """
     latest, modifications = dataset_tracker.since(key, timestamp)
-    if modifications or latest < 0 or timeout <= 0:
+    if modifications or latest < 0 or (timeout is not None and timeout <= 0):
         return latest, modifications
     try:
         await asyncio.wait_for(dataset_tracker.modified[key].wait(), timeout)
--- a/main.py
+++ b/main.py
@@ -580,7 +580,7 @@ async def get_dataset_modification(
         timeout: The timeout in seconds for awaiting new modifications.
           None for no timeout (wait until done), and 0 or negative for non-blocking.
           The actual execution time might exceed the timeout because of the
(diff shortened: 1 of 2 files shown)


kangz12345:
Oh, `asyncio.timeout()` is introduced in Python 3.11. I will update it.

kangz12345:
@BECATRUE 
I tested this PR! If you have further questions, please ask.

BECATRUE:
Please connect to the corresponding issue!

main.py
```diff
@@ -568,7 +568,7 @@ async def list_dataset() -> Tuple[str, ...]:
 
 @app.get("/dataset/master/modification/")
 async def get_dataset_modification(
-    key: str, timestamp: float
+    key: str, timestamp: float, timeout: Optional[float],
```
BECATRUE, line 571: How about setting the default value to `None`?

main.py
```diff
@@ -577,7 +577,19 @@ async def get_dataset_modification(
...
         key: The key of the target dataset.
         timestamp: The timestamp of the last update.
+        timeout: The timeout in seconds for awaiting new modifications.
+          None for no timeout (wait until done), and 0 or negative for non-blocking.
+          The actual execution time might exceed the timeout because of the
+          since() calls are not timed. Although the timeout becomes less precise,
+          it can guarantee a valid return value for even short timeouts.
     """
+    latest, modifications = dataset_tracker.since(key, timestamp)
+    if modifications or latest < 0 or timeout <= 0:
```
BECATRUE: The `timeout` may be `None`, so you should handle the case.

main.py
```diff
@@ -577,7 +577,19 @@ async def get_dataset_modification(
...
+        return latest, modifications
+    try:
+        await asyncio.wait_for(dataset_tracker.modified[key].wait(), timeout)
+    except asyncio.TimeoutError:
+        return latest, modifications
```
BECATRUE, line 592: I just want to check if I understood correctly.
In here, is it right that `latest` and `modifications` are `timestampt` and an empty tuple, respectively?

BECATRUE (changes requested):

main.py
kangz12345: Yes, you are right. Which indicates that there have been no modification.

main.py
kangz12345: I think `None` is _not_ recommended since it can block infinitely.
I didn't give the default to encourage setting a proper timeout.

BECATRUE (approved):
LGTM!

Issue #93, opened 2023-10-18
Manage datasets in `artiq-proxy` and return it immediately for "get" requests


kangz12345:
### Feature you want to implement
Currently, the `artiq-proxy` server does not manage the whole datasets, so `get_master_dataset()` performs RPC to get the artiq master's datasets.
However, this implementation has a synchronization issue: some modifications could be dropped or duplicated.
To resolve this issue, I will manage datasets in `artiq-proxy` itself and return it directly for get-dataset requests.

### How the feature is implemented
`get_master_dataset()` does not perform RPC to artiq master. Instead, it returns immediately the dataset of itself.

PR #95, opened 2023-10-19
Return timestamp with the dataset

### LINKED ISSUE #87: Use `Subscriber` to _watch_ dataset modifications and provide dataset fetching API

kangz12345:
This closes #87.

Note that this part will be updated by #93.

### DIFF
--- a/main.py
+++ b/main.py
@@ -547,17 +547,18 @@ async def list_result_directory() -> List[int]:
 
 
 @app.get("/dataset/master/")
-async def get_master_dataset(key: str) -> Union[int, float, List]:
-    """Returns the dataset broadcast to artiq master.
+async def get_master_dataset(key: str) -> Tuple[float, Union[int, float, List]]:
+    """Returns the current timestamp and the dataset broadcast to artiq master.
     
     Args:
         key: The key of the target dataset.
     """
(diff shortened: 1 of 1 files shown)


kangz12345:
@BECATRUE Please check this out!

BECATRUE:
I'm sorry for late check🥲 I will check this right now.

BECATRUE (approved):
👍

PR #99, opened 2023-10-25
Use built-in type annotation

### LINKED ISSUE #89: Use built-ins for type annotation
Since this repo support from Python 3.9+, we can use built-in types directly for generic type annotations, e.g., `list[str]` instead of `List[str]`.
Since this repo is not that big, so it seems okay to update the code to use the built-ins directly.

kangz12345:
This closes #89.

### DIFF
--- a/dataset.py
+++ b/dataset.py
@@ -6,7 +6,7 @@ import itertools
 import logging
 import time
 from collections import deque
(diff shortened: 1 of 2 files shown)


BECATRUE (approved):
LGTM!

BTW, I'm sorry we can't use `|` instead of `Union` because it is supported since Python 3.10.

PR #104, opened 2023-11-01
Fetch schedule in long-polling through `Subscriber`

### LINKED ISSUE #102: Plan for fetching scheduled queue through `Subscriber`
Like fetching the master dataset implemented in #96, I will updated the mechanism fetching the scheduled queue using `Subscriber`.

1. Implement the schedule tracker in `schedule.py`.
It is similar to `DatasetTracker`, but we don't need to track only modification, but just all scheduled queue, hence the implementation will be simpler.
3. Update `get_scheduled_queue()` in `main.py`. Now, we will read the tracked data from scheduler tracker.

BECATRUE:
This closes #102.

Now, the mechanism fetching the current schedule is implemented using `Subscriber`, similar to fetching datasets.
In addition, I replaced the term, "scheduled queue" with just "schedule".

I tested this with an actual ARTIQ device.


main.py
```diff
@@ -37,30 +37,7 @@
...
-        Args:
-            queue: A new scheduled queue.
-        """
-        self.updated_time = time.time()
-        self.queue = queue
-
-
-latest_schedule = ScheduleInfo()
-
+schedule_tracker: Optional[schd.ScheduleTracker] = None
```
kangz12345, line 38: How about writing these global assignments without empty lines between each line, just like the imports?

main.py
```diff
@@ -218,29 +201,23 @@ async def get_experiment_info(file: str) -> Any:
     return remote.examine(file)
 
 
-@app.get("/experiment/queue/", response_model=ScheduleInfo)
-async def get_scheduled_queue(updated_time: Optional[float] = None) -> Any:
-    """Gets the scheduled queue and returns it.
-
+@app.get("/schedule/")
+async def get_schedule(timestamp: float, timeout: Optional[float]) -> tuple[float, schd.Schedule]:
```
kangz12345, line 209: Please write the `Returns:` section.

schedule.py
```diff
@@ -0,0 +1,42 @@
...
+    """
+
+    def __init__(self):
+        """Extended."""
+        super().__init__()
+        self.modifed = asyncio.Event()
+        self.latest = -1
+
+    def get(self) -> tuple[float, Schedule]:
+        """Returns the current timestamp and schedule."""
```
kangz12345: "the current timestamp" looks awkward since it is actually the timestamp of the last modification.
If you don't want to repeat the definition of `latest`, you may write e.g. `Returns the latest timestamp and the current schedule`.

schedule.py
```diff
@@ -0,0 +1,42 @@
...
+        
+        All the coroutines that are waiting for the modified event will be awakened.
+        """
+        self.latest = time.time()
+        self.modifed.set()
+        self.modifed.clear()
+
+
+def notify_callback(tracker: ScheduleTracker, _mod: dict[str, Any]):
+    """Notifies modification to the tracker called as notify_cb() of sipyco system."""
```
kangz12345: Please write the `Args:` section.

tracker.py
```diff
@@ -0,0 +1,24 @@
+"""Module for base tracker."""
+
+from typing import Generic, Optional, TypeVar
+
+T = TypeVar("T")
+
+class Tracker(Generic[T]):
```
kangz12345, line 7: How about moving `notify_callback()` functions into `Tracker`? Their first arguments are the `Tracker` objects, so it would be fine to make them an abstract method of `Tracker`.
Then we don't have to use `functools.partial()`.

kangz12345 (changes requested):

tracker.py
BECATRUE: It looks much better! I will apply it.

kangz12345 (approved):
LGTM

PR #108, opened 2023-11-29
Support class name in submission

### LINKED ISSUE #107: Support class name in submission
Now, to simplify writing an experiment code, we use an inheritence among experiment classes.
For example,
```python
# .../test_library.py

from library.ttl import TTLWave, TTLPlay

class TestLibrary(TTLWave, TTLPlay):
    """Test a library"""

    def build(self):
        self.setattr_device("core")
        TTLWave.build(self)
        TTLPlay.build(self)

    @kernel
    def run(self):
        self.core.reset()
        self.wave(count=5, duration=1*us)
        delay(2*us)
        self.play()


# .../library/ttl.py

class TTLWave(EnvExperiment):
    """Create a long TTL wave"""

    def build(self):
        self.ttl = self.get_device("ttl22")

    @kernel
    def wave(self, count, duration):
        for i in range(count):
            self.ttl.pulse(duration / 2)
            delay(duration / 2)


class TTLPlay(EnvExperiment):
    """Play with 2 TTLs"""

    def build(self):
        self.ttl0 = self.get_device("ttl22")
        self.ttl1 = self.get_device("ttl23")

    @kernel
    def play(self):
        with parallel:
            with sequential:
                for _ in range(5):
                    self.ttl0.pulse(0.4*us)
                    delay(0.1*us)

            with sequential:
                for _ in range(2):
                    self.ttl1.pulse(1*us)
                    delay(1*us)
```

However, in this case, we CANNOT submit the experiment `test_library.py` with the original way;
`artiq_client submit .../test_library.py`,
because we SHOULD designate a specific experiment class as there are various classes, e.g., `TTLWave`, `TTLPlay`.

Hence, I will support to designate class name, i.e., [-c](https://m-labs.hk/artiq/manual/management_system.html#Named%20Arguments_repeat1) argument.
After done, I will update IQUIP as well as.

Like [artiq_client.py](https://github.com/m-labs/artiq/blob/master/artiq/frontend/artiq_client.py#L151), it looks sufficient just to add the class name into `expid`.

BECATRUE:
This closes #107. For details, please read the issue.

Now, we can submit an experiemtn with a specific class name.
```python
curl -X GET 'http://127.0.0.1:8000/experiment/submit/?file=test_jaehun/test_library_inheritence.py&cls=TestLibrary'
```

### DIFF
--- a/main.py
+++ b/main.py
@@ -327,17 +327,18 @@ def write_vcd(self):
 @app.get("/experiment/submit/")
 async def submit_experiment(  # pylint: disable=too-many-arguments, too-many-locals
     file: str,
+    cls: Optional[str] = None,
     args: str = "{}",
     pipeline: str = "main",
     priority: int = 0,
     timed: Optional[str] = None,
     visualize: bool = False,
-    cls: str = ""
 ) -> int:
     """Submits the given experiment file.
     
     Args:
         file: The path of the experiment file.
+        cls: The class name of the experiment to be submitted.
         args: The arguments to submit which must be a JSON string of a dictionary.
           Each key is an argument name and its value is the value of the argument.
         pipeline: The pipeline to run the experiment in.
@@ -346,8 +347,7 @@ async def submit_experiment(  # pylint: disable=too-many-arguments, too-many-loc
           None for no due date.
         visualize: If True, the experiment file is modified for visualization.
           The original file and vcd file are saved in the visualize path set in config.json.
-        cls: The class name of the experiment.
-          It is only necessary when the visualize argument is True.
+          The cls argument should be given.
     
     Returns:
         The run identifier, an integer which is incremented at each experiment submission.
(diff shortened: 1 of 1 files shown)


main.py
```diff
@@ -346,8 +347,7 @@ async def submit_experiment(  # pylint: disable=too-many-arguments, too-many-loc
           None for no due date.
         visualize: If True, the experiment file is modified for visualization.
           The original file and vcd file are saved in the visualize path set in config.json.
-        cls: The class name of the experiment.
-          It is only necessary when the visualize argument is True.
+          The cls argument should be given.
```
kangz12345, line 350: Why is this line required?

BECATRUE: It is because `cls` is necessary for visualization.
However, as we decided to use DAX, the visualizing feature will be deleted, so if you think it is unnecessary, I will remove it.

kangz12345: Oh, so you mean the `cls` should be given if `visualize` is `True`, right?

BECATRUE: Yes! You're right.

kangz12345 (approved):

PR #110, opened 2023-12-06
Lifespan is not working in Nix OS

### LINKED ISSUE #109: Lifespan is not working in Nix OS
**Describe the bug**
For safe management of various versions, I tested to use Nix OS configuration. 
In this process, I found that `app = FastAPI(lifespan=lifespan)` is not working.
I tried to realize the exact cause, but I couldn't..

However, as a temporary solution, we can use `startup` and `shutdown` event, the previous version of `lifespan`.

BECATRUE:
This resolves #109.

### DIFF
--- a/main.py
+++ b/main.py
@@ -9,7 +9,6 @@ import os
 import posixpath
 import shutil
 import time
-from contextlib import asynccontextmanager
 from datetime import datetime
 from enum import Enum
 from typing import Any, Optional, Union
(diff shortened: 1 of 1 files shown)


kangz12345:
Is this still preferred even if we are not using Nix OS anymore?

BECATRUE:
Oh, I forgot closing this issue.
Thanks for checking it.

PR #114, opened 2024-01-04
Send schedule using web socket

### LINKED ISSUE #113: Send schedule through web socket
As a part of #112, I will send schedule to each client through web socket.

Using a web socket, each client only need to connect to the proxy server once.
The server sends the updated schedule whenever it is modified, and clients always wait until a packet is received.

BECATRUE:
Now, a client i.e. IQUIP can fetch schedule using a web socket.

For test, run the code below.
```python
import json

def main():
    uri = "ws://localhost:8000/schedule/"

        while True:
            response = websocket.recv()
            data = json.loads(response)
            print(data)

main()
```

Its result is the following.

image

Additionally, I found out we didn't check Pylint for other files except for `main.py`, thus I updated it.

This closes #113.

### DIFF
--- a/main.py
+++ b/main.py
@@ -221,7 +221,7 @@ async def get_schedule(websocket: WebSocket):
         schedule = schedule_tracker.get()
         await websocket.send_json(schedule)
         while True:
-            await asyncio.wait_for(schedule_tracker.modifed.wait(), None)
+            await schedule_tracker.modifed.wait()
             schedule = schedule_tracker.get()
             await websocket.send_json(schedule)
     except websockets.exceptions.ConnectionClosedError:
--- a/main.py
+++ b/main.py
@@ -17,8 +17,9 @@ from typing import Any, Optional, Union
 import h5py
 import numpy as np
 import pydantic
+import websockets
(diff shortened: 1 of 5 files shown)


main.py
```diff
@@ -205,26 +206,28 @@ async def get_experiment_info(file: str) -> Any:
...
+    await websocket.accept()
     try:
-        await asyncio.wait_for(schedule_tracker.modifed.wait(), timeout)
-    except asyncio.TimeoutError:
-        return latest, schedule
-    return schedule_tracker.get()
+        schedule = schedule_tracker.get()
+        await websocket.send_json(schedule)
+        while True:
+            await asyncio.wait_for(schedule_tracker.modifed.wait(), None)
```
kangz12345: Isn't it equivalent to:
```python
await schedule_tracker.modified.wait()
```

kangz12345 (changes requested):
`WebSocket` looks so cool!

main.py
BECATRUE: Wow... Fantastic😮

kangz12345 (approved):
LGTM

PR #116, opened 2024-01-09
Send dataset using websocket

### LINKED ISSUE #115: Send dataset through web socket
As the last part of #112, I will send dataset information through web socket, like #113.

- `get_master_dataset`: Remove `timestamp`.
- `list_dataset()`: Replace with web socket. (for future dataset explorer)
- `get_datset_modification`: Remove `timestamp` and `timeout`, and replace with web socket.

BECATRUE:
Now, the proxy server supports to send dataset in real-time through a web socket.

### `get_master_dataset()`: GET /dataset/master/
It will not be used, but I left it for a convenience, e.g., debugging.

### `list_dataset()`: Web socket /dataset/master/list/
It sends the available dataset in real-time.
An example experiment code is like below.
```python
import time

class DatasetTest(EnvExperiment):
    """Dataset Test"""

    def build(self):
        self.setattr_device("core")

    @kernel
    def run(self):
        self.set_dataset("a", [], broadcast=True)
        for i in range(10):
            self.append_to_dataset("a", i+1)
            time.sleep(1)
        self.set_dataset("b", [], broadcast=True)
        time.sleep(2)
```

First, start the proxy sever and run the following web socket client code.
```python
import json

def main():
    uri = "ws://localhost:8000/dataset/master/list/"

        for response in websocket:
            response = json.loads(response)
            print(response)
         
main()
```
Then, submit the experiment code.

The result of the client code is as follows.

image

### `get_dataset_modification()`: Web socket /dataset/master/modification/
Like sending the dataset list, it sends the modification of the specific dataset in real-time.
In addition, it sends at least a second apart to prevent high network demand.

First, start the proxy server and submit the above experiment.
```python
import json

def main():
    uri = "ws://localhost:8000/dataset/master/modification/"

        key = json.dumps("a")
        websocket.send(key)
        for response in websocket:
            response = json.loads(response)
            if response is None:
                break
            print(response)
            
main()
```

Finally, run the client code above and the result is as follows.

image
The first three lines are the current dataset, parameters, and units, respectively.
And, the following lines are modifications of the dataset.

This closes #115.

### DIFF
--- a/dataset.py
+++ b/dataset.py
@@ -73,6 +73,7 @@ class DatasetTracker(Tracker[Dataset]):
     """Holds dataset modifications and provides searching API.
     
     Attributes:
+        list_modified: The event set when the dataset list is modified.
         modified: Dict(key=dataset_name, value=event) where the event is set
           when any new modification to the dataset is added.
     """
@@ -83,6 +84,7 @@ class DatasetTracker(Tracker[Dataset]):
             maxlen: The maximum length of modification queues.
         """
         super().__init__()
+        self.list_modified = asyncio.Event()
         self.modified: dict[str, asyncio.Event] = {}
         self._maxlen = maxlen
         self._modifications: dict[str, ModificationQueue] = {}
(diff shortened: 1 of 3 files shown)


BECATRUE:
To pass GitHub actions, I need to fix config files for actions.
However, I already modified them in #114, hence I will push this PR again after #114 is merged.

BECATRUE:
After #114 is merged, I will rebase this branch and reopen it.

PR #117, opened 2024-01-18
Send dataset using websocket

### LINKED ISSUE #115: Send dataset through web socket

BECATRUE:
Now, the proxy server supports to send dataset in real-time through a web socket.

### `get_master_dataset()`: GET /dataset/master/
It will not be used, but I left it for a convenience, e.g., debugging.

### `list_dataset()`: Web socket /dataset/master/list/
It sends the available dataset in real-time.
An example experiment code is like below.
```python
import time

class DatasetTest(EnvExperiment):
    """Dataset Test"""

    def build(self):
        self.setattr_device("core")

    @kernel
    def run(self):
        self.set_dataset("a", [], broadcast=True)
        for i in range(10):
            self.append_to_dataset("a", i+1)
            time.sleep(1)
        self.set_dataset("b", [], broadcast=True)
        time.sleep(2)
```

First, start the proxy sever and run the following web socket client code.
```python
import json

def main():
    uri = "ws://localhost:8000/dataset/master/list/"

        for response in websocket:
            response = json.loads(response)
            print(response)
         
main()
```
Then, submit the experiment code.

The result of the client code is as follows.

image

### `get_dataset_modification()`: Web socket /dataset/master/modification/
Like sending the dataset list, it sends the modification of the specific dataset in real-time.
In addition, it sends at least a second apart to prevent high network demand.

First, start the proxy server and submit the above experiment.
```python
import json

def main():
    uri = "ws://localhost:8000/dataset/master/modification/"

        key = json.dumps("a")
        websocket.send(key)
        for response in websocket:
            response = json.loads(response)
            if response is None:
                break
            print(response)
            
main()
```

Finally, run the client code above and the result is as follows.

image
The first three lines are the current dataset, parameters, and units, respectively.
And, the following lines are modifications of the dataset.

This closes #115.

### DIFF
--- a/main.py
+++ b/main.py
@@ -562,7 +562,6 @@ async def list_dataset(websocket: WebSocket):
     await websocket.accept()
     try:
         datasets = dataset_tracker.datasets()
-        # send the current dataset list on the first connection.
         await websocket.send_json(datasets)
         while True:
             await asyncio.wait_for(dataset_tracker.list_modified.wait(), None)
--- a/main.py
+++ b/main.py
@@ -549,10 +549,26 @@ async def get_master_dataset(key: str) -> Union[int, float, list]:
     return data
 
 
-@app.get("/dataset/master/list/")
-async def list_dataset() -> tuple[str, ...]:
-    """Returns the list of datasets available in artiq master."""
-    return dataset_tracker.datasets()
+@app.websocket("/dataset/master/list/")
+async def list_dataset(websocket: WebSocket):
+    """Sends the list of datasets available in artiq master whenever it is modified.
+    
+    Args:
+        websocket: The web socket object.
+    """
+    await websocket.accept()
+    try:
+        datasets = dataset_tracker.datasets()
+        # send the current dataset list on the first connection.
+        await websocket.send_json(datasets)
+        while True:
+            await asyncio.wait_for(dataset_tracker.list_modifed.wait(), None)
+            datasets = dataset_tracker.datasets()
+            await websocket.send_json(datasets)
+    except websockets.exceptions.ConnectionClosedError:
+        logger.info("The connection for sending the dataset list is closed.")
+    except websockets.exceptions.WebSocketException:
+        logger.exception("Failed to send the dataset list.")
 
 
 @app.get("/dataset/master/modification/")
--- a/main.py
+++ b/main.py
@@ -145,9 +145,9 @@ async def lifespan(_app: FastAPI):
     load_device_db()
     _schedule_task = await init_schedule_tracker()
     _dataset_task = await init_dataset_tracker()
-    await connect_moninj()
+    # await connect_moninj()
     yield
-    await mi_connection.close()
+    # await mi_connection.close()
 
 
 app = FastAPI(lifespan=lifespan)
@@ -589,8 +589,8 @@ async def get_dataset_modification(websocket: WebSocket):
     await websocket.accept()
     try:
         name = await websocket.receive_json()
-        timestamp, data = dataset_tracker.get(name)
-        await websocket.send_json(data)
+        timestamp, dataset = dataset_tracker.get(name)
+        await websocket.send_json(dataset)
         _, parameters = dataset_tracker.get(f"{name}.parameters")
         await websocket.send_json(parameters)
         _, units = dataset_tracker.get(f"{name}.units")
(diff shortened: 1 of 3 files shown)


main.py
```diff
@@ -534,52 +534,80 @@ async def list_result_directory() -> list[int]:
...
-    """Returns the current timestamp and the dataset broadcast to artiq master.
+async def get_master_dataset(key: str) -> Union[int, float, list]:
+    """Returns the dataset broadcast to artiq master.
 
     Args:
         key: The key of the target dataset.
 
     Returns:
-        If the dataset is not initialized or does not exist, it returns (-1, ()).
+        If the dataset is not initialized or does not exist, it returns an empty tuple.
```
kangz12345, line 544: Oh, the previous docstring was wrong: `(-1, [])` as `data` is not a tuple but a list.

main.py
```diff
@@ -534,52 +534,80 @@ async def list_result_directory() -> list[int]:
...
 
+    Args:
+        websocket: The web socket object.
+    """
+    await websocket.accept()
+    try:
+        datasets = dataset_tracker.datasets()
+        await websocket.send_json(datasets)
+        while True:
+            await asyncio.wait_for(dataset_tracker.list_modified.wait(), None)
```
kangz12345: I guess this is equivalent to
```python
await dataset_tracker.list_modified.wait()
```

main.py
```diff
@@ -534,52 +534,80 @@ async def list_result_directory() -> list[int]:
...
+    After accepted, it sends the current dataset list immediately.
+    Then, it sends the dataset list every time it is modified.
 
+    Args:
+        websocket: The web socket object.
+    """
+    await websocket.accept()
+    try:
+        datasets = dataset_tracker.datasets()
+        await websocket.send_json(datasets)
```
kangz12345: Since `datasets` is not used in any other places and the method name `datasets()` is clear enough, I think you can just write it in a single line:
```python
await websocket.send_json(dataset_tracker.datasets())`
```

main.py
```diff
@@ -534,52 +534,80 @@ async def list_result_directory() -> list[int]:
...
+        while True:
+            await asyncio.wait_for(dataset_tracker.list_modified.wait(), None)
+            datasets = dataset_tracker.datasets()
+            await websocket.send_json(datasets)
```
kangz12345: Here as well, `datasets` is not used in any other places.

main.py
```diff
@@ -534,52 +534,80 @@ async def list_result_directory() -> list[int]:
...
-    return dataset_tracker.since(key, timestamp)
+        name = await websocket.receive_json()
+        timestamp, dataset = dataset_tracker.get(name)
+        await websocket.send_json(dataset)
+        _, parameters = dataset_tracker.get(f"{name}.parameters")
+        await websocket.send_json(parameters)
+        _, units = dataset_tracker.get(f"{name}.units")
+        await websocket.send_json(units)
+        while True:
+            await asyncio.wait_for(dataset_tracker.modified[name].wait(), None)
```
kangz12345: You don't have to use `wait_for()`.

main.py
```diff
@@ -534,52 +534,80 @@ async def list_result_directory() -> list[int]:
...
+            latest, modifications = dataset_tracker.since(name, timestamp)
+            if latest < 0:  # dataset is overwritten or removed
+                await websocket.send_json(None)
+                break
+            if latest > timestamp + 1:  # changes within 1 second are buffered
+                timestamp = latest
+                await websocket.send_json(modifications)
```
kangz12345: This loop runs too busy even though it does not send anything within 1 second.
How about inserting a [`sleep()`](https://docs.python.org/3/library/asyncio-task.html#asyncio.sleep) after sending modifications, instead of checking the timestamp?

kangz12345 (changes requested):

main.py
BECATRUE: Umm.. I think the previous docstring was correct since `DatasetTracker.get()` might return `(-1, ())`.
Instead, the type hinting in terms of this function looks wrong: `Union[int, float, list, tuple]`.

main.py
BECATRUE: Why couldn't I think of this simple way?😮 I will apply your idea.

main.py
kangz12345: Oh I thought FastAPI automatically casts tuples to lists because of the return type annotation.
However, I'm not sure about that..
I'm okay with updating the type hint.

kangz12345 (approved):
LGTM

PR #120, opened 2024-01-26
Synchronize TTL status

### LINKED ISSUE #119: Synchronize TTL status
I will implement to synchronize TTL status, i.e., whether overriding and level are on or off.
In addition, I will support to turn on or off all TTLs.

#### Synchronize TTL status
- Once the server is started, a `CommMonInj` connection is established and monitoring `TTLProbe.level.value`, `TTLOverride.en.value`, and  `TTLOverride.level.value` is started.
- Once its callback function is called, an asyncio event corresponding to each TTL monitoring signal is set.
- This is connected to an IQUIP client by a websocket, and it waits the asyncio events above.
- Once it wakes up, it sends the modification to the IQUIP client.
- ~~To know the status of the first connection, I will use a trick: After the connection is made, a short time later, through [get_injection_status()](https://github.com/m-labs/artiq/blob/master/artiq/coredevice/comm_moninj.py#L77) the callback function is called. It may have the initial information.~~
- To know the status of the first connection, once the connection is established, it sends the current value.

#### Turn on or off all TTLs
- Implement `async def set_all_ttl_levels(value: bool)` similar to `set_ttl_level()`.

BECATRUE:
For details, please refer to #119.

Yet, I couldn't test whether it works well.. because this should be tested with the actual ARTIQ hardware.
I'll test it as soon as possible.

This closes #119.


main.py
```diff
@@ -33,11 +34,90 @@
...
+    @dataclasses.dataclass
+    class Status:
+        """TTL status.
+        
+        Fields:
+            value: The current value.
+            modified: The asyncio event set once the value is updated.
+        """
+        value: Optional[bool] = None
+        modified: asyncio.Event = asyncio.Event()
```
kangz12345, line 63: The default value will be created once and all the instances will share it as the default value. Is this behavior intended? If not, consider using `default_factory` argument in [`field()`](https://docs.python.org/3/library/dataclasses.html#dataclasses.field).

main.py
```diff
@@ -33,11 +34,90 @@
...
+            self.value = bool(value)
+            self.modified.set()
+            self.modified.clear()
+
+    def __init__(self):
+        """Extended."""
+        self.connection = CommMonInj(self.monitor_cb, self.injection_status_cb)
+        self.outputs: dict[int, MonInj.Status] = {}
+        self.levels: dict[int, MonInj.Status] = {}
+        self.overriding = MonInj.Status()
```
kangz12345, line 80: `overriding` is not a dict since we are currently using a single override status for all TTLs?

main.py
```diff
@@ -33,11 +34,90 @@
...
+            ty: The type of injection status. See artiq.coredevice.comm_moninj.TTLOverride.
+            value: The updated injection status.
+        """
+        if ty == TTLOverride.level.value:
+            self.levels[channel].update(value)
+        elif ty == TTLOverride.en.value:
+            self.overriding.update(value)
+
+
+mi: Optional[CommMonInj] = None
```
kangz12345, line 118: I guess it should be `Optional[MonInj]`.

kangz12345 (changes requested):

main.py
BECATRUE: I did not intend it..
I think as we discussed, even if all the instances share it, it can detect a chance (although quite inefficient), hence it seemed to work well.

BECATRUE:
I will implement this feature using a queue instead of lots of events.
Thus, I close this PR.

PR #128, opened 2024-03-18
Fetch dataset list corresponding RID

### LINKED ISSUE #125: Fetch dataset list corresponding RID
To the next step of #124, I will implement a function to fetch dataset list corresponding a specific RID.

It reads the corresponding H5 result file referred to #49.
And it returns the dataset list stored in the result file.

### Additional context
It is a part of snu-quiqcl/iquip#258.

BECATRUE:
This closes #125.

For test, write a command like below.
```shell
$ curl -X GET locathost:8000/dataset/rid/list/?rid=123

['test_a', 'test_a.parameters', 'test_a.units']
```

### DIFF
--- a/main.py
+++ b/main.py
@@ -583,6 +583,17 @@ async def get_master_dataset(key: str) -> Union[int, float, list, tuple]:
     return data
 
 
+@app.get("/dataset/rid/list/")
+async def list_dataset_from_rid(rid: int) -> list[str]:
+    result_dir_path = posixpath.join(configs["master_path"], configs["result_path"])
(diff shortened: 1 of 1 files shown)


kangz12345 (approved):
I wonder how fast the `glob` will find the RID file, but I believe it will be fast enough. I think this task does not require very high performance.
However, if we find it becomes slower as we generate much more data files, we might consider improving the algorithm later.
(I guess you did not find such inconvenience during the test!)

BECATRUE:
Honestly, I didn't consider which one between `glob` and `os` is better. A simple search shows that `os` is still faster than `glob`.
But, it wasn't enough to make me feel slow and I think in terms of speed, networking is more dominative than searching files.

PR #129, opened 2024-03-19
Fetch dataset from result file

### LINKED ISSUE #126: Fetch dataset from result file
To the next step of #125, I will implement a function to fetch a specific dataset from the given RID.

It fetches a specific dataset based on the given RID and dataset name.

### Additional context
It is a part of snu-quiqcl/iquip#258.

BECATRUE:
This closes #126.

For test, write a command like below.
```shell
$ curl -X GET "localhost:8000/dataset/rid/?rid=759&key=test_a"

[[0,0],[1,1],[2,2],[3,3],[4,4],[5,5],[6,6],[7,7],[8,8],[9,9]]
```

### DIFF
--- a/main.py
+++ b/main.py
@@ -568,8 +568,19 @@ async def list_rid_from_date_hour(date: str, hour: Optional[int] = None) -> list
 
 
 @app.get("/dataset/rid/")
-async def get_rid_dataset(rid: int, key: str) -> Union[int, float, list]:
-    pass
+async def get_rid_dataset(rid: int, key: str) -> Optional[Union[int, float, list]]:
+    result_dir_path = posixpath.join(configs["master_path"], configs["result_path"])
+    result_file_path = posixpath.join(result_dir_path, "*", "*", f"{str(rid).zfill(9)}*.h5")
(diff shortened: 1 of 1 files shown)


main.py
```diff
@@ -567,6 +567,31 @@ async def list_rid_from_date_hour(date: str, hour: Optional[int] = None) -> list
...
+        key: Target dataset key.
+    
+    Returns:
+        If the dataset does not exist, it returns None.
+    """
+    result_dir_path = posixpath.join(configs["master_path"], configs["result_path"])
+    result_file_path = posixpath.join(result_dir_path, "*", "*", f"{str(rid).zfill(9)}*.h5")
+    result_file_list = glob.glob(result_file_path)
+    if len(result_file_list) != 1:
+        return None
```
kangz12345: How about making it as a function? I remember the same code was introduced in #125.

kangz12345 (changes requested):

main.py
BECATRUE: I see. I will separate it as a function.

kangz12345 (approved):
LGTM

PR #131, opened 2024-03-22
Synchronize TTL status using queue

### LINKED ISSUE #130: Synchronize TTL status using queue
According to the discussion in #120, I will re-implement to synchronize TTL status using queue in order to reduce the burden of many `asyncio.Event`s.

The overall process is similar to fetching dataset modifications in `get_dataset_modification()`:
1. Receive the TTL channels the IQUIP client subscribes to.
2. Send the current status.
3. Run a loop.
3-1. Get the current status.
3-2. If not updated, wait for the event set when the queue is updated.
3-3. Send the fetched status.
3-4. Sleep a second.

The TTL connection manager like `MonInj` class in "BECATRUE/119/synchronize-ttl-old" branch has the following properties:
- Attributes
  - A `SortedQueue` with modifications containing only the channel and the type.
  - A Dictionary whose keys are the channel and the type, and values are the latest values.
  - An event set when any value is updated.
- Methods
  - Return the current status (for initial connection).
  - Return the modifications since a certain timestamp.

BECATRUE:
This closes #130.

As discussed in #120, I implemented to synchronize TTL status through queue instead of many events.

Here is what I did:
- Make a TTL manager, `MonInj`.
- Change to handle overrides separately, not together.
- Support for controlling one or more TTLs in a single request.

For test, please refer the corresponding IQUIP PR.


main.py
```diff
@@ -19,26 +22,170 @@
...
 
 configs = {}
 device_db = {}
 
-mi_connection: Optional[CommMonInj] = None
-
 dataset_tracker: Optional[dset.DatasetTracker] = None
 schedule_tracker: Optional[schd.ScheduleTracker] = None
 
+class MonInj:
```
kangz12345: Could you explain why you designed the class like this, defining class-variables and a class-method?
Isn't it better to separate this as another module and put them out of this class?

BECATRUE: Umm.. I never thought about modularization😅
I just wanted to handle TTL connections through a single object, like `CommMonInj` in ARTIQ.

If you think modularization is better, I will do that.

BECATRUE: @kangz12345 Can you check this reply?

kangz12345: I think the use of `MonInj` looks like that of a module.
The enum classes are using `MonInj` as a namespace.
If there is no reason to use a class, I think creating a new module would be better.
You can freely express your opinions.

BECATRUE:
I made the TTL manager as a separate module.
Please check it. @kangz12345

ttl.py
```diff
@@ -0,0 +1,153 @@
...
+    """Maps TTL devices and channels.
+    
+    Args:
+        ttl_devices: See main.configs.
+        device_db: See main.device_db.
+    """
+    for device in ttl_devices:
+        channel = device_db[device]["arguments"]["channel"]
+        device_to_channel[device] = channel
+        channel_to_device[channel] = device
```
kangz12345: I think this can be packed within a class, e.g., `DeviceChannelMapping`, which provides methods `device(channel: int)` and `channel(device: str)`.
Then `map_device_channel()` would be the constructor of the class.
This design has a benefit that many mappings can exist and it does not use global (module-level) objects.

ttl.py
```diff
@@ -0,0 +1,153 @@
...
+class MonitorType(enum.Enum):
+    """Monitoring value type."""
+    PROBE = "probe"
+    LEVEL = "level"
+    OVERRIDE = "override"
+
+
+@dataclasses.dataclass
+class StatusType:
+    """Monitoring status type.
```
kangz12345, line 56: I don't get the meaning of "monitoring status".
This is merely a combination of a channel and a monitor type.
Could you explain about the intuition?

ttl.py
```diff
@@ -0,0 +1,153 @@
...
+    channel: int
+    monitor_type: MonitorType
+
+    def __hash__(self) -> int:
+        """Overridden."""
+        return hash(f"{str(self.channel)}_{self.monitor_type.value}")
+
+
+# {"MonitorType value": {"TTL device name": "modified value"}}
+Modifications = dict[str, dict[str, bool]]
```
kangz12345, line 71: Isn't it `dict[MonitorType, dict[str, bool]]` then?

ttl.py
```diff
@@ -0,0 +1,153 @@
...
+            timestamp: Timestamp of the latest update.
+        """
+        modifications = {ty.value: {} for ty in MonitorType}
+        latest, modification_types = self.queue.tail(timestamp)
+        filtered_modification_types = filter(
+            lambda ty: channel_to_device[ty.channel] in devices, set(modification_types)
+        )
+        for ty in filtered_modification_types:
+            device = channel_to_device[ty.channel]
+            modifications[ty.monitor_type.value][device] = self.values[ty]
```
kangz12345: In the for loop, it needs the `device` anyway. How about not filtering and checking it in the for loop, using `continue`?

main.py
```diff
@@ -727,33 +730,78 @@ async def get_result(rid: str, result_file_type: ResultFileType) -> FileResponse
...
+        websocket: The web socket object.
+    """
+    await websocket.accept()
+    try:
+        devices = await websocket.receive_json()
+        latest, status = ttl_manager.current_status(devices)
+        await websocket.send_json(status)
+        while True:
+            latest, modifications = ttl_manager.modifications_since(devices, latest)
+            if not sum(len(m) for m in modifications.values()):  # no modification
```
kangz12345: How about `if not any(modifications.values())`?

kangz12345 (changes requested):

ttl.py
BECATRUE: I have thought what is the best class name, but I couldn't come up with the best one..
I intended that a status type is an individual type we monitor, so I wrote such that.
However, as you said, it looks vague so how about just "Status type mapping to a value"?

ttl.py
BECATRUE: Umm.. A key is the value of `MonitorType`, so I set it as `str`. Isn't it correct?

ttl.py
BECATRUE: It can reduce the duplicated for loop! I will apply it.

main.py
BECATRUE: Wow, I didn't come up with this idea! Thank you👍

ttl.py
kangz12345: I see, then why not using `MonitorType` itself instead of its `value`?

BECATRUE: It is because `Modifications` is the type of data it sends to a client, so I defined it to a type which can be JSONized.

kangz12345: Oh I see, thank you for the explanation!

ttl.py
kangz12345: I got your point. Thank you for the explanation.

kangz12345 (approved):
LGTM

PR #133, opened 2024-03-26
Fix bug that last data is missed

### LINKED ISSUE #132: Last data is missed when fetching dataset
### Describe the bug
This problem is issued in snu-quiqcl/iquip#261.

Similar to #130, I will try to solve it.

BECATRUE:
This closes #132.

I fixed the bug as follows:
1. After sleep, it fetches the dataset modifications first.
2. If there is no modification, it waits the `modified` event. After the event is set, it goes to the start point of the loop.
3. Otherwise, it sends the appropriate data to the client.

### DIFF
--- a/main.py
+++ b/main.py
@@ -682,11 +682,13 @@ async def get_dataset_modification(websocket: WebSocket):
         _, units = dataset_tracker.get(f"{name}.units")
         await websocket.send_json(units)
         while True:
-            await dataset_tracker.modified[name].wait()
             latest, modifications = dataset_tracker.since(name, latest)
             if latest < 0:  # dataset is overwritten or removed
                 await websocket.send_json(None)
                 break
+            if not modifications:  # no modification:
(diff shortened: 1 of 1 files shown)


main.py
```diff
@@ -146,9 +146,9 @@ async def lifespan(_app: FastAPI):
     load_device_db()
     _schedule_task = await init_schedule_tracker()
     _dataset_task = await init_dataset_tracker()
-    await connect_moninj()
+    # await connect_moninj()
     yield
-    await mi_connection.close()
+    # await mi_connection.close()
```
kangz12345: Maybe these should be reverted

kangz12345 (changes requested):

main.py
BECATRUE: Oh, I'm sorry for the mistake. I just reverted it.

kangz12345 (approved):
LGTM

PR #138, opened 2024-04-09
Use `requirements.txt` instead of direct installation

### LINKED ISSUE #137: Use `requirements.txt` instead of direct installation
To manage easily, we need to use a requirements file instead of direct installation in GitHub actions.

BECATRUE:
This closes #137.


kangz12345 (approved):
Now we don't use `h5py` and `numpy`, right?

BECATRUE:
No! We still use these libraries, but they are included in `artiq` library.

PR #142, opened 2024-04-10
Update config file to include only general information

### LINKED ISSUE #141: Enable to set config file path when starting server
> Since this is a repo for general use, the system specific information should not be included. Also it would be better for security.
Let's make the config template just like iquip.

_Originally posted by @kangz12345 in https://github.com/snu-quiqcl/artiq-proxy/issues/140#issuecomment-2046373979_

Like IQUIP, I will enable to set the config file path in the command to start the server.

In FastAPI, [Setting](https://fastapi.tiangolo.com/advanced/settings/?h=#pydantic-settings) is supported to enable to set a config outside of the Python code.
Using this, I will add an enviroment variable, `config_path`, the target config file path.

BECATRUE:
This closes #141.

### DIFF
--- a/config.json
+++ b/config.json
@@ -6,13 +6,9 @@
     "core_addr": "192.168.1.75",
     "master_addr": "localhost",
     "notify_port": 3250,
-    "ttl_devices": [
-        "ttl4", "ttl5", "ttl6", "ttl7", "ttl8", "ttl9", "ttl10", "ttl11", "ttl12", "ttl13",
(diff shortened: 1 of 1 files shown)


kangz12345 (changes requested):
Let's make this `config_template.json` and add `config.json` in the gitignore just like `iquip`.

BECATRUE:
Oh, I see! Then, I should find how to start a shell script with arguments so I will update it tomorrow.

BECATRUE:
The contents of #141 was changed, so I will open another PR, instead of updating this.

PR #143, opened 2024-04-11
Enable to set config file path when starting server

### LINKED ISSUE #141: Enable to set config file path when starting server

BECATRUE:
_(updated)_

This closes #141.

Now, one can set the target config file path as follows:
```shell
$ CONFIG_PATH="config.json" ./server.sh           # target config file = "config.json"

$ CONFIG_PATH="config_template.json" ./server.sh  # target config file = "config_template.json"

$ ./server.sh                                     # target config file = default ("config.json")
```

For convenience, the command line arguments are supported in `server.sh`:
```shell
$ ./server.sh -c config.json           # target config file = "config.json"

$ ./server.sh -c config_template.json  # target config file = "config_template.json"

$ ./server.sh                          # target config file = default ("config.json")
```

The followings are example results:


kangz12345:
Great! `BaseSetting` would enable more flexible parameterization.
As you said, however, setting the environment variables is not familiar to our usual cases.
How about exploiting the shell script to make it as a command line argument?
In other words, for example,
```shell
./server.sh -c config.json
```
and for its implementation, [a Stack Overflow answer](https://unix.stackexchange.com/a/505342) may help.

BECATRUE:
I see! It looks much helpful. I will implement such the shell script.

BECATRUE:
Please check the updated `server.sh`! @kangz12345 
(The re-require button has been disabled😅)

server.sh
```diff
@@ -1 +1,23 @@
...
+while getopts "c:" opt
+do
+    case "$opt" in
+        c) config_path=$OPTARG ;;
+        *) echo "Invalid option" ; helpFunction ;;
+    esac
+done
+
+if [ -z "$config_path" ]; then
+    python -m uvicorn main:app --host 0.0.0.0 --port 8000
```
kangz12345: How about defining variables for host address and port number in the shell script? For convenient modification of the script.

kangz12345 (changes requested):

kangz12345 (approved):
LGTM!

PR #149, opened 2025-02-24
Sequencer compatible config

miinukim:


kangz12345:
Please provide more information for us to understand the pull request.
For example, the goal or motivation of this change, and a brief explanation about how you implemented the desired features.

miinukim:
Merging to the 'develop' branch is not required, as it will be used to control sequencer drivers and thus much of the original features will not be used. Thus, maintenance alongside the develop branch is not necessary.

